PR #9, opened 2022-03-31
Bump google.golang.org/api

Fedosin:
To include API related to pd-balanced disks, we have to bump "google.golang.org/api" library.

Generated by:
$ go get -u "google.golang.org/api"
$ go mod vendor


alexander-demichev (approved):
/lgtm

PR #11, opened 2022-06-01
Bug 2087039: update dependencies to K8s 1.24, go 1.18

damdo:
- Update the dependencies to K8s 1.24.
- Bump project to go 1.18.


alexander-demichev (approved):
/approve

hack/fetch-ext-bins.sh
```diff
@@ -26,7 +26,7 @@ if [ -n "$TRACE" ]; then
   set -x
 fi
 
-k8s_version=1.22.1
+k8s_version=1.23.5
```
lobziik, line 29: should it be 1.24.something?

lobziik: ok, no kubebuilder for 1.24 yet, ignore this

damdo: We always lag behind several versions with those envtest kubebuilder tools (On some repos we where even on 1.19). As they release them significantly later unfortunately. That's why we were still on 1.22 on the previous release. For now we are stuck on 1.23.x.

PR #14, opened 2022-07-26
Add windows machine support

elmiko:
this change adds detection logic to the create machine function of the
reconciler that will look at the machine labels to determine if the new
machine will be a windows instance. if so, it will then attempt to
locate the bootscript in either the provider spec or a secret in the
namespace. if neither is found an error will be returned, otherwise the
machine will be created with the script placed into the GCP metadata for
the instance.


elmiko:
i'd like to add a few more tests to this

elmiko:
going to need to update this once, https://github.com/openshift/machine-api-operator/pull/1049 merges

elmiko:
i have updated this and am removing the WIP status.

this should be good to go with the WMCO as it stands today.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -20,12 +21,15 @@ import (
...
-	machineTypeFmt      = "zones/%s/machineTypes/%s"
-	acceleratorTypeFmt  = "zones/%s/acceleratorTypes/%s"
+	userDataSecretKey            = "userData"
+	requeueAfterSeconds          = 20
+	instanceLinkFmt              = "https://www.googleapis.com/compute/v1/projects/%s/zones/%s/instances/%s"
+	kmsKeyNameFmt                = "projects/%s/locations/%s/keyRings/%s/cryptoKeys/%s"
+	machineTypeFmt               = "zones/%s/machineTypes/%s"
+	acceleratorTypeFmt           = "zones/%s/acceleratorTypes/%s"
+	windowsScriptMetadataKey     = "sysprep-specialize-script-ps1"
+	windowsScriptSecretName      = "windows-user-data"
```
jrvaldes, line 31: Ideally, this should be an exported constant in [windows](https://github.com/openshift/machine-api-provider-gcp/pull/14/commits/f22bd895a9d859c1168f9f82c0a47e4e13a9ac8c#diff-359c93fbc08f569c3abf79881f0b426e681c46a5e76765d05e921f16fdb3bf99R13)

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -261,6 +265,47 @@ func (r *Reconciler) create() error {
...
+			if !found {
+				// if we did not find this key in the previous metadata, add it now
+				instance.Metadata.Items = append(instance.Metadata.Items, &compute.MetadataItems{
+					Key:   windowsScriptMetadataKey,
+					Value: &script,
+				})
+			}
+		} else {
+			// we didn't find a script
+			return machinecontroller.CreateMachine("cannot find Windows script in providerSpec or Secret")
```
jrvaldes, line 304: For my own education; does `machinecontroller.CreateMachine()` throws an exception/error?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -414,6 +459,39 @@ func (r *Reconciler) getCustomUserData() (string, error) {
...
+		if item.Key == windowsScriptMetadataKey {
+			script = windows.RemovePowershellTags(*item.Value)
+			break
+		}
+	}
+
+	return script
+}
+
+func (r *Reconciler) getWindowsScriptFromSecret() (string, error) {
```
jrvaldes, line 477: Consider documenting this function.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -414,6 +459,39 @@ func (r *Reconciler) getCustomUserData() (string, error) {
...
+	var scriptSecret corev1.Secret
+	if err := r.coreClient.Get(context.Background(), client.ObjectKey{Namespace: windowsScriptSecretNamespace, Name: windowsScriptSecretName}, &scriptSecret); err != nil {
+		if apimachineryerrors.IsNotFound(err) {
+			return "", machinecontroller.InvalidMachineConfiguration("Windows script data secret %q in namespace %q not found: %v", windowsScriptSecretName, windowsScriptSecretNamespace, err)
+		}
+		return "", fmt.Errorf("error getting Windows script data secret %q in namespace %q: %v", windowsScriptSecretName, windowsScriptSecretNamespace, err)
+	}
+
+	data, exists := scriptSecret.Data[windowsScriptMetadataKey]
```
jrvaldes, line 486: The `sysprep-specialize-script-ps1` key may not be available if the script was loaded from a secret [created](https://github.com/openshift/windows-machine-config-operator/blob/4c09463d252e2bb20e133a41372be1d438267ba2/pkg/secrets/secrets.go#L54) by WMCO.

jrvaldes (commented):
@elmiko thanks for working on this, added a few comments. PTAL

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -20,12 +21,15 @@ import (
...
-	kmsKeyNameFmt       = "projects/%s/locations/%s/keyRings/%s/cryptoKeys/%s"
-	machineTypeFmt      = "zones/%s/machineTypes/%s"
-	acceleratorTypeFmt  = "zones/%s/acceleratorTypes/%s"
+	userDataSecretKey            = "userData"
+	requeueAfterSeconds          = 20
+	instanceLinkFmt              = "https://www.googleapis.com/compute/v1/projects/%s/zones/%s/instances/%s"
+	kmsKeyNameFmt                = "projects/%s/locations/%s/keyRings/%s/cryptoKeys/%s"
+	machineTypeFmt               = "zones/%s/machineTypes/%s"
+	acceleratorTypeFmt           = "zones/%s/acceleratorTypes/%s"
+	windowsScriptMetadataKey     = "sysprep-specialize-script-ps1"
```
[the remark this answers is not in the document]
aravindhp: @jrvaldes `sysprep-specialize-script-ps1` is mandated by GCP. PTAL [Metadata keys for Windows startup scripts](https://cloud.google.com/compute/docs/instances/startup-scripts/windows#metadata-keys).

pkg/cloud/gcp/actuators/machine/reconciler.go
aravindhp: That's correct. WMCO is agnostic to what GCP expects for it's meta-data key.

aravindhp (commented):
Thanks for helping out with this @elmiko. Mostly LGTM.

pkg/cloud/gcp/actuators/machine/reconciler.go
jrvaldes: I would say; Yes [100%](https://github.com/openshift/windows-machine-config-operator/blob/4c09463d252e2bb20e133a41372be1d438267ba2/pkg/secrets/secrets.go#L54) and defer to @openshift/openshift-team-windows-containers in case I'm missing something.

sebsoto: > Are we providing any guarantee around the userData key being used in WMCO right now?

 If you're asking about a guarantee to users, we have no such guarantee. We maintain the userdata secret and do not allow users to change it. The user provides a private key in the form of another secret that they have control over, which is used when creating the userdata secret

JoelSpeed: I want to know that if we are reading from the `userData` map key, that that will always work going forward and that it's not likely to be changed to something else by WMCO, I guess the key name forms part of the API contract at this point right?

sebsoto: @JoelSpeed I don't see a need to ever change it, the key is what it is solely to interact with the Machine API. WMCO will not be changing the key name unless what is expected by MAPI is changed.

jrvaldes: +1

pkg/cloud/gcp/actuators/machine/reconciler.go
sebsoto: Does this need to be a constant at all?
I would think you'd just want to use the value of `userDataSecret` in the machine spec.

From [the OCP Windows docs](https://docs.openshift.com/container-platform/4.10/windows_containers/creating_windows_machinesets/creating-windows-machineset-aws.html#windows-machineset-aws_creating-windows-machineset-aws), we expect/require users to use the `windows-user-data` secret.

pkg/cloud/gcp/actuators/machine/reconciler.go
elmiko: @jrvaldes that function does wrap the creation of an `Error` type. it has an unfortunate name =(

pkg/cloud/gcp/actuators/machine/reconciler.go
elmiko: i can update this

pkg/cloud/gcp/actuators/machine/reconciler.go
elmiko: will do

pkg/cloud/gcp/actuators/machine/reconciler.go
elmiko: i'll update this

pkg/cloud/gcp/actuators/machine/reconciler.go
elmiko: oh wait, actually, @sebsoto , i think i used the value from the WMCO because what if we have a situation where the user needs to actually specify a `userDataSecret` in addition to whatever the WMCO is creating?

elmiko:
updated
* fix secret access to use `userData` key
* add docstring to `getWindowsScriptFromSecret`

i left the one conversation about getting the secret name from the provider spec open, i don't think we want to do this but we should discuss a little more.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -676,3 +700,91 @@ func TestSetMachineCloudProviderSpecifics(t *testing.T) {
...
+	}{
+		{
+			name: "Valid secret",
+			secret: &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      windowsScriptSecretName,
+					Namespace: defaultNamespace,
+				},
+				Data: map[string][]byte{
+					windowsScriptMetadataKey: []byte(secretBlob),
```
JoelSpeed, line 737: I think this should be `wmcoSecretDataKey`

elmiko: oops XD

elmiko:
updated
* fix unit tests

JoelSpeed:
/approve

Please get one of the WMCO team to add an LGTM if they are happy

pkg/cloud/gcp/actuators/machine/reconciler.go
sebsoto: > what if we have a situation where the user needs to actually specify a userDataSecret in addition to whatever the WMCO is creating

Currently thats not something we support, but it could be nice to have the flexibility to do that in the future if desired

sebsoto:
I've tried out one of your earlier revisions and was able to add a Windows node with the sysprep given via the userdata :tada: 
Thanks @elmiko, changes LGTM, but adding a hold in case you want https://github.com/openshift/machine-api-operator/pull/1055 to merge first.
/hold

elmiko:
thanks @sebsoto , i'll figure out what we're doing with the other patch and make sure we get this merged. appreciate all the help =)

elmiko:
there is some slowdown getting my PR merged into MAO with another helper function, it's merely a small upgrade here. i am going to remove the hold so we can get on to testing.
/hold cancel

PR #16, opened 2022-09-13
update windows script handling

elmiko:
this change removes the code that would look specifically for the windows powershell secret. instead it relies on the normal user data from the `userDataSecret` in conjuction with the windows o/s label to inject the windows script in the proper metadata key.

this change does preserve the behavior for the user to specify values in
the metadata which will override the values from the secret.


JoelSpeed:
Changes LGTM but QE didn't get it to work, let's make sure they're happy before we merge (may need to make this a bug)

elmiko:
+1, i've left some comments with QE. need to work out where the failure happened.

elmiko:
it looks like this is working as expected, see [comment from qe on jira](https://issues.redhat.com/browse/OCPCLOUD-1568?focusedCommentId=20952374&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-20952374)

elmiko:
/hold

found a bug that needs addressing

JoelSpeed (commented):
/lgtm

PR #19, opened 2022-10-03
Updating ose-machine-api-provider-gcp images to be consistent with ART

lobziik:
Cherry-picked from #18 

Apply go fmt in follow up commit


JoelSpeed:
/override ci/prow/e2e-gcp-operator 

On the latest run, all tests passed but the teardown failed

PR #20, opened 2022-11-04
OCPCLOUD-1755: Register control plane machines to instance group

RadekManak:
Control plane machines on GCP need to be in instance group for the internal load balancer to work. This PR automatically assigns the instance to an already created instance group when it is being updated and removes it from instance group when it is being deleted.


JoelSpeed (commented):
Changes look good, @RadekManak can you please explain what manual testing you've done to verify this is working as expected?

JoelSpeed:
/approve

Testing discussion happened on the Jira, QE have approved

PR #21, opened 2022-12-01
OCPBUGS-1557: refactor restartPolicyToBool function

elmiko:
to be more thorough about checking for empty string and preemptible instances. previously this function only checked if the value was equal to "Always". but, this misses the default omission case where we should be defaulting to always restart but weren't. additionally, when the instance is preemptible it is not possible to set a restart policy of always.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -2,6 +2,7 @@ package machine
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -66,9 +67,21 @@ func containsString(sli []string, str string) bool {
 	return false
 }
 
-func restartPolicyToBool(policy machinev1.GCPRestartPolicyType) *bool {
-	restart := policy == machinev1.RestartPolicyAlways
-	return &restart
+func restartPolicyToBool(policy machinev1.GCPRestartPolicyType, preemptible bool) (*bool, error) {
+	// for more information about how the restart policy works, see the GCP docs at
+	// https://cloud.google.com/compute/docs/instances/setting-vm-host-options#settingoptions
+	if len(policy) == 0 {
+		return nil, nil
+	} else if policy == machinev1.RestartPolicyAlways {
+		if preemptible {
+			return nil, errors.New("preemptible instances cannot be automatically restarted")
+		}
+		return pointer.Bool(true), nil
+	} else if policy == machinev1.RestartPolicyNever {
+		return pointer.Bool(false), nil
+	}
+
+	return nil, fmt.Errorf("unrecognized restart policy: %s", policy)
 }
 
 // machineTypeAcceleratorCount represents nvidia-tesla-A100 GPUs which are only compatible with A2 machine family
@@ -161,9 +174,14 @@ func (r *Reconciler) create() error {
 		},
 		Scheduling: &compute.Scheduling{
 			Preemptible:       r.providerSpec.Preemptible,
-			AutomaticRestart:  restartPolicyToBool(r.providerSpec.RestartPolicy),
 			OnHostMaintenance: string(r.providerSpec.OnHostMaintenance),
 		},
+	}
+
+	if automaticRestart, err := restartPolicyToBool(r.providerSpec.RestartPolicy, r.providerSpec.Preemptible); err != nil {
+		return machinecontroller.InvalidMachineConfiguration("failed to determine restart policy: %v", err)
+	} else {
+		instance.Scheduling.AutomaticRestart = automaticRestart
 	}
 
 	var guestAccelerators = []*compute.AcceleratorConfig{}
--- a/pkg/cloud/gcp/actuators/machine/reconciler_test.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler_test.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/openshift/api/machine/v1beta1"
 	machinev1 "github.com/openshift/api/machine/v1beta1"
 	machinecontroller "github.com/openshift/machine-api-operator/pkg/controller/machine"
 	computeservice "github.com/openshift/machine-api-provider-gcp/pkg/cloud/gcp/actuators/services/compute"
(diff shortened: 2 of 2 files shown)


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -66,8 +66,22 @@ func containsString(sli []string, str string) bool {
...
-func restartPolicyToBool(policy machinev1.GCPRestartPolicyType) *bool {
-	restart := policy == machinev1.RestartPolicyAlways
+func restartPolicyToBool(policy machinev1.GCPRestartPolicyType, preemptible bool) *bool {
+	// for more information about how the restart policy works, see the GCP docs at
+	// https://cloud.google.com/compute/docs/instances/setting-vm-host-options#settingoptions
+	restart := false
+
+	if preemptible {
+		// preemptible instances cannot be set to always restart
+		klog.V(4).Info("Setting preemptible instance restart policy to never")
```
elmiko, line 76: i added this log line because we are potentially deviating from what the user has specified in the providerSpec. it is possible that a user might have set the restart policy to "Always", which is not usable with preemptible instances. i figure we should at least log the change, although this might be something we want to add the validation webhook for gcp.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -66,8 +66,22 @@ func containsString(sli []string, str string) bool {
...
+		return &restart
+	}
+
+	if policy == machinev1.RestartPolicyAlways || len(policy) == 0 {
+		// if the policy is unset (zero value), we use Always as the default as specified by the GCP docs
+		restart = true
```
wking, line 82: Do we need to inline GCP's upstream choice?  From [GCP's docs][1]:

> Compute Engine configures all VMs with default values unless you specify otherwise.

I'd expect that to be something like:

```go
if len(policy) == 0 {
  return nil, nil // caller had no opinion, so let the cloud provider pick a default
} else if policy == machinev1.RestartPolicyAlways {
  if preemptible {
    return nil, errors.New("preemptible instances cannot be automatically restarted") // caller can complain about this with a Machine reference, so the log-reading user can follow the trail of the impossible request.
  }
  return pointer.Bool(true) // caller requested automatic restarts
} else if policy == machinev1.RestartPolicyNever {
  return pointer.Bool(false) // caller requested manual restarts
}
return nil, fmt.Errorf("unrecognized restart policy %q", policy) // caller can complain about this with a Machine reference, so the log-reading user can follow the trail of the impossible request.
```

or something?

[1]: https://cloud.google.com/compute/docs/instances/host-maintenance-options#settingoptions

elmiko: hmm, yeah, perhaps i went too deep replicating the logic. i guess we can pass through the `nil` values when it is not specified and allow the platform to make the decision.

elmiko:
i refactored the check to be more similar to what gcp is doing, also added some errors so that we can see how the failures have occurred.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -330,6 +330,62 @@ func TestCreate(t *testing.T) {
...
+				RestartPolicy: machinev1.RestartPolicyNever,
+			},
+			validateInstance: func(t *testing.T, instance *compute.Instance) {
+				if instance.Scheduling != nil && instance.Scheduling.AutomaticRestart != nil && *instance.Scheduling.AutomaticRestart != false {
+					t.Errorf("Expected to find instance automatic restart set to true")
+				}
+			},
+		},
+		{
+			name: "Restart policy is set to Never when using preemptible instances",
```
elmiko, line 364: this test case might be worth dropping since it's not necessarily testing the logic in our function, but it does confirm the default behavior coming out of the gcp sdk. happy to remove if folks don't think we should include it.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -330,6 +330,62 @@ func TestCreate(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "Restart policy defaults to Always when not set",
+			validateInstance: func(t *testing.T, instance *compute.Instance) {
+				if instance.Scheduling != nil && instance.Scheduling.AutomaticRestart != nil && *instance.Scheduling.AutomaticRestart != true {
+					t.Errorf("Expected to find instance automatic restart set to true")
```
wking, line 337: nit: the `name` is stale; it should be "...to nil when not set...".

And the final `*instance.Scheduling.AutomaticRestart != true` seems to allow `AutomaticRestart == nil` and `AutomaticRestart == &true`, while we only want to allow the former.  Can we use:

```go
if instance.Scheduling != nil && instance.Scheduling.AutomaticRestart != nil {
```

for "it's fine if you have no `Scheduling` opinions.  Or if you have scheduling opinions but no `AutomaticRestart` opinions"?

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -330,6 +330,62 @@ func TestCreate(t *testing.T) {
...
+				}
+			},
+		},
+		{
+			name: "Restart policy is set to Always when set",
+			providerSpec: &machinev1.GCPMachineProviderSpec{
+				RestartPolicy: machinev1.RestartPolicyAlways,
+			},
+			validateInstance: func(t *testing.T, instance *compute.Instance) {
+				if instance.Scheduling != nil && instance.Scheduling.AutomaticRestart != nil && *instance.Scheduling.AutomaticRestart != true {
```
wking, line 347: In this case, `Scheduling == nil` should be fatal, so we want something like:

```go
if instance.Scheduling == nil || instance.Scheduling.AutomaticRestart == nil || *instance.Scheduling.AutomaticRestart != true {
```

pkg/cloud/gcp/actuators/machine/reconciler_test.go
elmiko: i think i'm going to refactor these tests so that we are just testing the functionality of the new code i am adding. this is starting to feel like we are testing the gcp sdk and i don't think that is appropriate since we don't control the sdk.

JoelSpeed:
Code changes look reasonable, let me know when you get the tests together and I'll give them a review too

elmiko:
i have removed the previous unit tests in favor of a suite that is focused on the `restartPolicyToBool` function.

elmiko:
i added some tests for the reconciler to check that we are returning the proper error conditions.

elmiko:
one more minor cleanup on the printed error messages from the tests.

elmiko:
added a couple reconciler tests to check that the affirmative cases do not produce errors.

JoelSpeed (commented):
/lgtm

PR #22, opened 2022-12-02
OCPBUGS-3904: Register unknown instance groups

odvarkadaniel:
This PR solves a bug, where machines were stuck in add/delete phase when removing a failure domain from CPMS.
Whenever we encounter an instance group that is unknown to us, we register it.

Edit: Have to add that new instance group to load balancer.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -577,17 +577,24 @@ func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) er
 	return nil
 }
 
+// ensureInstanceGroup ensures that the instance group exists.
+// If the instance group doesn't exist, we try and register it and also assign
+// it to a backend service correctly.
 func (r *Reconciler) ensureInstanceGroup(instanceGroupName string) error {
-	if err := r.checkInstanceGroupExists(instanceGroupName); err != nil {
+	// Get an instance group so we can check that it does in fact exist
+	_, err := r.computeService.InstanceGroupGet(r.projectID, r.providerSpec.Zone, instanceGroupName)
+	if isNotFoundError(err) {
 		// Handle the creation of a new instance group
 		if err := r.registerNewInstanceGroup(); err != nil {
 			return fmt.Errorf("failed to register the new instance group named %s: %v", instanceGroupName, err)
 		}
+	} else if err != nil {
+		return fmt.Errorf("instanceGroupGet request failed: %v", err)
 	}
 
 	registered, err := r.checkRegistrationOfBackend()
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to retrieve the backend service: %v", err)
 	}
 
 	if !registered {
@@ -597,15 +604,6 @@ func (r *Reconciler) ensureInstanceGroup(instanceGroupName string) error {
 		}
 	}
 
-	return nil
-}
-
-// checkInstanceGroupExists checks whether an instancegroup exists.
-func (r *Reconciler) checkInstanceGroupExists(instanceGroupName string) error {
-	_, err := r.computeService.InstanceGroupGet(r.projectID, r.providerSpec.Zone, instanceGroupName)
-	if err != nil {
-		return fmt.Errorf("instanceGroupGet request failed: %v", err)
-	}
 	return nil
 }
 
@@ -674,7 +672,7 @@ func (r *Reconciler) registerInstanceToControlPlaneInstanceGroup() error {
 	instanceGroupName := r.controlPlaneGroupName()
 
 	if err := r.ensureInstanceGroup(instanceGroupName); err != nil {
-		return fmt.Errorf("failed to ... : %v", err)
+		return fmt.Errorf("failed to ensure that instance group %s is a proper instance group: %v", instanceGroupName, err)
 	}
 
 	instanceSets, err := r.fetchRunningInstancesInInstanceGroup(r.projectID, r.providerSpec.Zone, instanceGroupName)
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -635,7 +635,10 @@ func (r *Reconciler) fetchRunningInstancesInInstanceGroup(projectID string, zone
 		},
 	)
 	if err != nil {
-		return nil, fmt.Errorf("instanceGroupsListInstances request failed: %v", err)
+		if errRegister := r.registerNewInstanceGroup(); errRegister != nil {
+			return nil, fmt.Errorf("instanceGroupsListInstances request failed: %v, failed to register the new instance group named %v: %v", err, instaceGroup, errRegister)
+		}
+		return nil, fmt.Errorf("instanceGroupsListInstances request failed: %v, registered the new instance group named %v successfully", err, instaceGroup)
 	}
 
 	instanceSets := sets.NewString()
@@ -644,6 +647,21 @@ func (r *Reconciler) fetchRunningInstancesInInstanceGroup(projectID string, zone
 	}
 
 	return instanceSets, nil
+}
+
+// registerNewInstanceGroup registers an instance group when there is an instance
+// that is using that unkown instance group.
+func (r *Reconciler) registerNewInstanceGroup() error {
+	_, err := r.computeService.InstanceGroupInsert(r.projectID, r.providerSpec.Zone, &compute.InstanceGroup{
+		Name:   r.controlPlaneGroupName(),
+		Region: r.providerSpec.Region,
+		Zone:   r.providerSpec.Zone,
+	})
+	if err != nil {
+		return fmt.Errorf("instanceGroupInsert request failed: %v", err)
+	}
+
+	return nil
 }
 
 // ControlPlaneGroupName generates the name of the instance group that this instace should belong to.
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -577,8 +577,31 @@ func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) er
 	return nil
 }
 
-// ensureInstanceGroupExists checks whether an instancegroup exists.
-func (r *Reconciler) ensureInstanceGroupExists(instanceGroupName string) error {
+func (r *Reconciler) ensureInstanceGroup(instanceGroupName string) error {
+	if err := r.checkInstanceGroupExists(instanceGroupName); err != nil {
+		// Handle the creation of a new instance group
+		if err := r.registerNewInstanceGroup(); err != nil {
+			return fmt.Errorf("failed to register the new instance group named %s: %v", instanceGroupName, err)
+		}
+	}
+
+	registered, err := r.checkRegistrationOfBackend()
+	if err != nil {
+		return err
+	}
+
+	if !registered {
+		// Handle the registration of backend to backend service
+		if err := r.updateBackendServiceWithInstanceGroup(); err != nil {
+			return fmt.Errorf("failed to update the backend service with new instance group %s: %v", instanceGroupName, err)
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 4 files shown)


odvarkadaniel:
When a failure domain got deleted/added, a machine used to get stuck in provisioning/deleting phase.  There used to be an issue with a instance group not being registered. When that happens now, we register that instance group and then assign it to a backend service. There is an issue with the new instance group now being ready in time to be assigned to the backend service, we solve that by wrapping the function that does that in `retry.OnError()`. We set a backoff for that and amount of repetitions. This seems to be working nicely, but I wonder, if there is some nicer solution to that?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -635,7 +637,13 @@ func (r *Reconciler) fetchRunningInstancesInInstanceGroup(projectID string, zone
 		},
 	)
 	if err != nil {
-		return nil, fmt.Errorf("instanceGroupsListInstances request failed: %v", err)
+		if errRegister := r.registerNewInstanceGroup(); errRegister != nil {
+			return nil, fmt.Errorf("instanceGroupsListInstances request failed: %v, failed to register the new instance group named %v: %v", err, instaceGroup, errRegister)
```
JoelSpeed, line 641: Instance group is a string right? Use `%s` please

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -635,7 +637,13 @@ func (r *Reconciler) fetchRunningInstancesInInstanceGroup(projectID string, zone
...
+		}
+		if errBackendService := r.updateBackendServiceWithInstanceGroup(); errBackendService != nil {
```
JoelSpeed, line 643: Use %s for the instance group substitution if it's a string, else try and get the name of the instance group from the object

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -646,6 +654,82 @@ func (r *Reconciler) fetchRunningInstancesInInstanceGroup(projectID string, zone
 	return instanceSets, nil
 }
 
+// registerNewInstanceGroup registers an instance group when there is an instance
+// that is using that unkown instance group. After the creation of the instance group
+// we add patch our internal load balancer with that new instance group.
```
JoelSpeed, line 659:
```suggestion
// we patch our internal load balancer with that new instance group.
```

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -646,6 +654,82 @@ func (r *Reconciler) fetchRunningInstancesInInstanceGroup(projectID string, zone
...
+}
+
+func (r *Reconciler) updateBackendServiceWithInstanceGroup() error {
+	// We have to wait for the instance group to be ready
+	backoff := wait.Backoff{
+		Duration: 10 * time.Second,
+		Steps:    10,
+		Factor:   1.0,
+		Jitter:   1.0,
+	}
```
JoelSpeed, line 682: If you return an error from the function, it causes the controller to requeue and do backoff anyway, we shouldn't bother with our own back off here as this makes this blocking, requeue, process other things, come back to it

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -646,6 +654,82 @@ func (r *Reconciler) fetchRunningInstancesInInstanceGroup(projectID string, zone
...
+// FQDNInstanceGroup generates a FQDN for our instance group.
+// It is neccessary for the addition of the instance group to the backend service.
+func (r *Reconciler) FQDNInstanceGroup() string {
+	return fmt.Sprintf("https://www.googleapis.com/compute/v1/projects/%s/zones/%s/instanceGroups/%s", r.projectID, r.providerSpec.Zone, r.controlPlaneGroupName())
+}
+
+// backendServiceName generates the name of a cluster's backend service
+func (r *Reconciler) backendServiceName() string {
+	return fmt.Sprintf("%s-api-internal", r.machine.Labels[machinev1.MachineClusterIDLabel])
+}
```
JoelSpeed, line 721: Is the name of this always true? I wonder if in non-ipi installs this could vary, is there any way to look this up rather than assume it?

pkg/cloud/gcp/actuators/machine/reconciler.go
odvarkadaniel: I experienced that, if I used `retry.DefaultBackoff` it didn't provide enough time for the instancegroup to become ready.

odvarkadaniel: Also, the issue is with the structure of the code. If we don't patch the backend service right after registering the new instance group, we never get to execute that code again.

pkg/cloud/gcp/actuators/machine/reconciler.go
odvarkadaniel: This didn't cross my mind at all. In my installments, it was always in this format. I have no clue whether I can find that information somewhere. We can see that here as well: [https://console.cloud.google.com/compute/instanceGroups/list?project=openshift-gce-devel&pageState=(%22instanceGroupsTable%22:(%22f%22:%22%255B%255D%22))](GCP instanceGroupTable)
I guess the internal backend service, that get created by the installment is always in that format, but the user can have it's custom backend service?

JoelSpeed: Lets leave as is for now and see how we get on, I think this should be ok as we only expect control plane machines to exist on clusters created by the installer

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -577,11 +577,48 @@ func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) er
...
+
+	return fmt.Errorf("couldn't find the instance group %s in a backend service", r.FQDNInstanceGroup())
+}
+
 // registerInstanceToControlPlaneInstanceGroup ensures that the instance is assigned to the control plane instance group of its zone.
 func (r *Reconciler) registerInstanceToControlPlaneInstanceGroup() error {
 	instanceSelfLink := fmtInstanceSelfLink(r.projectID, r.providerSpec.Zone, r.machine.Name)
 	instanceGroupName := r.controlPlaneGroupName()
 
+	if err := r.ensureInstanceGroupExists(instanceGroupName); err != nil {
```
JoelSpeed, line 610: Is there a way to distinguish between a not found error and any other error? 

Also, when we use `ensure` we mean it to be, check if the thing exists, create it if it doesn't. So you should move the `registerNewInstanceGroup` into the `ensureInstanceGroup` function

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -577,11 +577,48 @@ func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) er
...
+		if errRegister := r.registerNewInstanceGroup(); errRegister != nil {
+			return fmt.Errorf("failed to register the new instance group named %s: %v", instanceGroupName, errRegister)
+		}
+	}
+
+	if err := r.ensureInstanceGroupInBackendService(); err != nil {
+		if errPatch := r.updateBackendServiceWithInstanceGroup(); err != nil {
```
JoelSpeed, line 617: Same comments as above, but I'd probably expect to move the backend bit into `ensureInstanceGroup` as well so there's just a single new branch from this function

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -635,7 +672,7 @@ func (r *Reconciler) fetchRunningInstancesInInstanceGroup(projectID string, zone
 		},
 	)
 	if err != nil {
-		return nil, fmt.Errorf("instanceGroupsListInstances request failed: %v", err)
+		return nil, fmt.Errorf("instanceGroupsListInstances request failed: %v, registered the new instance group named %v successfully", err, instaceGroup)
```
JoelSpeed, line 675: I think this error should be changed back right?

JoelSpeed (commented):
So I think the `ensureInstanceGroup` function should look something like
```
func ensureInstanceGroup() error {
  if err := checkInstanceGroupExists(); IsNotFound(err) {
    // Handle creating new instance group
  } else if err != nil {
    return fmt.Errorf("genuine error", err)
  } 

  if registered, err := checkRegistrationOfBackend(); err != nil {
    return err
  }

  if !registered {
     // Handle registration of backend
  }
}
```

Then we can just call the `ensureInstanceGroup` function from the top function and we have a minimal logical diff in that top function

odvarkadaniel:
@JoelSpeed, Fixed tests & restructure the code as you mentioned above. Tested it and everything is working.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -577,11 +577,106 @@ func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) er
 	return nil
 }
 
+func (r *Reconciler) ensureInstanceGroup(instanceGroupName string) error {
+	if err := r.checkInstanceGroupExists(instanceGroupName); err != nil {
```
JoelSpeed, line 581: Does this error always mean we need to create a new instance group? Can we check the error for a 404 somehow please

odvarkadaniel:
Found about function that was already in the reconciler file - `isNotFoundError(err)`. Had to get rid of function that was wrapping method `compute.instanceGroup.Get` from the google API to use it - it was messing up the returning err type - `googleapi.Error` and we check for err code 404. Tests are fixed. Code should now be clean as well IMO. All the jobs should pass as well.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -577,11 +577,101 @@ func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) er
 	return nil
 }
 
+func (r *Reconciler) ensureInstanceGroup(instanceGroupName string) error {
```
JoelSpeed, line 580: Please add a godoc to explain what this function is doing

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -577,11 +577,101 @@ func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) er
...
+	// Get an instance group so we can check that it does in fact exist
+	_, err := r.computeService.InstanceGroupGet(r.projectID, r.providerSpec.Zone, instanceGroupName)
+	if isNotFoundError(err) {
+		// Handle the creation of a new instance group
+		if err := r.registerNewInstanceGroup(); err != nil {
+			return fmt.Errorf("failed to register the new instance group named %s: %v", instanceGroupName, err)
+		}
+	} else if err != nil {
+		return err
```
JoelSpeed, line 589: Make sure to wrap errors with additional context

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -577,11 +577,101 @@ func (r *Reconciler) processTargetPools(desired bool, poolFunc poolProcessor) er
...
+	}
+
+	registered, err := r.checkRegistrationOfBackend()
+	if err != nil {
+		return err
```
JoelSpeed, line 594: Make sure to wrap errors with additional context

odvarkadaniel: Done

pkg/cloud/gcp/actuators/machine/reconciler.go
odvarkadaniel: Done

pkg/cloud/gcp/actuators/machine/reconciler.go
odvarkadaniel: Done

odvarkadaniel:
The e2e GCP tests should now pass with this [PR](https://github.com/openshift/machine-api-operator/pull/1094) merged.

odvarkadaniel:
The test failing now, even after the introduction of the additional permission has to do something with the leakage of custom roles created by CCO in CI. Will try to run that e2e test again once this is solved.

PR #24, opened 2022-12-05
OCPBUGS-860: Set sync period for Machine controller

JoelSpeed:
On other providers we set the sync period to 10 minutes down from the default of 10 hours. This means we pick up changes within the cloud provider in a reasonable time frame.

For example, if a user adds an external IP address manually, this should be picked up so that CSRs for the serving certificates are approved in a timely manner, this isn't working well with the default sync period.

10 minutes is used in other providers such as AWS and Azure.

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -84,12 +84,17 @@ func main() {
 
 	cfg := config.GetConfigOrDie()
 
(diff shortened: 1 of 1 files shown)


Fedosin (approved):
/lgtm

Fedosin:
/hold

for e2e tests to pass

PR #32, opened 2022-12-12
[OCPBUGS-4574]: Machines with invalid zones stuck in no phase

odvarkadaniel:
This PR solves issue from this [JIRA ticket](https://issues.redhat.com/browse/OCPBUGS-4574).
If we create a new machine set with an invalid zone (that doesn't exist), we get a new machine, that is stuck in no phase, zone, region and machine type. If we find a machine with invalid zone, we want to put it in phase `Failed`.
Once we delete this machine set in in valid zone, we want this new machine to be deleted as well. We check for an invalid machine configuration and also check the phase in this case.


pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -96,7 +112,21 @@ func (a *Actuator) Exists(ctx context.Context, machine *machinev1.Machine) (bool
 	// When create()/update() try to store machineSpec/status this might result in
 	// "Operation cannot be fulfilled; the object has been modified; please apply your changes to the latest version and try again."
 	// Therefore we don't close the scope here and we only store spec/status atomically either in create()/update()"
-	return newReconciler(scope).exists()
+	_, exists, err := newReconciler(scope).exists()
+	if isInvalidMachineConfigurationError(err) {
```
JoelSpeed, line 116: The main machine controller should try and recognise this situation no? Does it not handle invalid configuration on exists? 

What would happen if there is an actual Machine there, the exists call returns false, but with an error, so we requeue? We only set the phase to failed when it's empty so this should be ok I guess

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -96,7 +112,21 @@ func (a *Actuator) Exists(ctx context.Context, machine *machinev1.Machine) (bool
...
+		// we have to make sure the phase goes as "Failed".
+		if machine.Status.Phase == nil {
+			machine.Status.Phase = pointer.String("Failed")
+		}
+		// If the machine has, e.g. invalid zone, and we delete the invalid machinset,
+		// we want to set the machine to "Deleting" phase and return nil as error.
+		// We need the error to be nil so we can successfully delete the invalid machine.
+		if *machine.Status.Phase == "Deleting" {
+			return false, nil
+		}
```
JoelSpeed, line 127: Hmm, this worries me, what if there is actually a Machine in the cluster, then we update the configuration to be invalid, say, the credentials expire, now we remove the Machine object, but the Machine itself remains in the cloud provider right?

If there's any chance of leaking machines we should probably close this as a won't fix

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -480,9 +480,9 @@ func validateMachine(machine machinev1.Machine, providerSpec machinev1.GCPMachin
 }
 
 // Returns true if machine exists.
-func (r *Reconciler) exists() (bool, error) {
+func (r *Reconciler) exists() (bool, bool, error) {
```
JoelSpeed, line 483: What is the new bool value representing? If this is just for the invalid zone, perhaps using an invalidmachineconfiguration error would be better served?

pkg/cloud/gcp/actuators/machine/actuator.go
odvarkadaniel: 1. I would think so as well, because of the comments that are spread in these sections of code, but unfortunately it doesn't recognize the situation.

2. With this fix, we distinguish whether we get a 404 error or just a regular error and on that basis, we determine what we want to do next. Either re-queue for a regular error, or if the machine doesn't exist (e.g. due to the configuration not being valid), we want to delete/fail that machine.

pkg/cloud/gcp/actuators/machine/actuator.go
odvarkadaniel: So are you talking about a machine being in phase `Running` and then we edit it, so it has an invalid configuration?
I have not tried that. Will investigate. Nevertheless, what you are saying makes sense.

odvarkadaniel: What we could theoretically do is: In function `exists()`, we would try and get that machine, if that failed, we would know that the machine was never created properly in the first place and then work with that information if it exists. What do you think?

pkg/cloud/gcp/actuators/machine/reconciler.go
odvarkadaniel: We need to know if, in this case the zone, doesn't exist. So we use that boolean to return the value to us so we can distinguish that case from regular errors.
And yes, the naming you are suggesting makes more sense.

odvarkadaniel:
@JoelSpeed, This new commit makes sure we don't leave any instances with invalid configs behind on GCP.
We check if a machine has a node assigned to it. If so, we know that this machine in fact exists and then we have to make sure to treat it properly. I tested this behavior by editing zone for a worker that was originally in zone `us-central-1a` and placed it into `us-central-1d`. First of all, if we find this invalid config for such machine, we Fail it. If we try to delete such machine, we are prompted by the logs, that we need to first make sure that this machine has valid configuration. Once we repair this configuration, we can successfully delete this machine.

pkg/cloud/gcp/actuators/machine/actuator.go
JoelSpeed: > So are you talking about a machine being in phase Running and then we edit it, so it has an invalid configuration?
I have not tried that. Will investigate. Nevertheless, what you are saying makes sense.

Yes, that's one of my concerns, if the VM exists, ie we have a provider ID set, then we must make sure, even if the configuration is no longer valid, that we don't remove the machine until we are certain the VM is gone

> we would try and get that machine, if that failed, we would know that the machine was never created properly in the first place and then work with that information if it exists. What do you think?

I think we can rely on the providerID being set, there are many reasons why a call to get the instance might fail, for example, if the user mistakenly changed the region or zone, we know the information in the Machine isn't right, but because there is a providerID, we know we did create a VM. 

By not removing the Machine in this case, we alert the user that some manual intervention is required, which I think is desired

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -96,7 +97,28 @@ func (a *Actuator) Exists(ctx context.Context, machine *machinev1.Machine) (bool
 	// When create()/update() try to store machineSpec/status this might result in
 	// "Operation cannot be fulfilled; the object has been modified; please apply your changes to the latest version and try again."
 	// Therefore we don't close the scope here and we only store spec/status atomically either in create()/update()"
-	return newReconciler(scope).exists()
+	invalidMachineConfiguration, exists, err := newReconciler(scope).exists()
+	if !invalidMachineConfiguration {
+		return exists, err
+	}
+
+	if exists {
```
JoelSpeed, line 105: Can this ever happen? Surely we don't know if the Machine exists when there is an invalid configuraiton?

pkg/cloud/gcp/actuators/machine/actuator.go
odvarkadaniel: Yes, the only thing that I had to do different is the part about `providerID`. Even machines that are not created properly have some. That's why I opted for node reference instead.

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -96,7 +97,28 @@ func (a *Actuator) Exists(ctx context.Context, machine *machinev1.Machine) (bool
...
+		machine.Status.Phase = pointer.String("Failed")
+	} else {
+		// If the machine has, e.g. invalid zone, and it doesn't have a phase set yet,
+		// we have to make sure the phase goes as "Failed".
+		if machine.Status.Phase == nil {
+			machine.Status.Phase = pointer.String("Failed")
+		}
```
JoelSpeed, line 112: Returning an invalid machine configuration error doesn't set the phase to failed?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -492,20 +492,26 @@ func (r *Reconciler) exists() (bool, error) {
 			// this error type bubbles back up to the machine-controller to allow
 			// us to delete machines that were never properly created due to
 			// invalid zone.
-			return false, machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("%s: Machine does not exist", r.machine.Name))
+
+			// If the machine has a node assigned, we know that it actually exists,
+			// but it also has an invalid configuration.
+			if r.machine.Status.NodeRef != nil {
+				return true, true, machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("%s: Zone does not exist", r.providerSpec.Zone))
```
JoelSpeed, line 499: I don't think we need this bool, you can unwrap/check that the error return type is an `invalidMachineConfiguration` within the parent scope, this would be more idiomatic

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -492,20 +492,26 @@ func (r *Reconciler) exists() (bool, error) {
 			// this error type bubbles back up to the machine-controller to allow
 			// us to delete machines that were never properly created due to
 			// invalid zone.
-			return false, machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("%s: Machine does not exist", r.machine.Name))
+
+			// If the machine has a node assigned, we know that it actually exists,
+			// but it also has an invalid configuration.
+			if r.machine.Status.NodeRef != nil {
```
JoelSpeed, line 498: NodeRef isn't good enough, NodeRef only comes once the node has joined the cluster. If you use providerID instead, this will be set as soon as the VM is created on the cloud provider, which is earlier than the NodeRef.

There is a period between the VM being created and the providerID being set that the current code would cause an issue

pkg/cloud/gcp/actuators/machine/actuator.go
JoelSpeed: >  Even machines that are not created properly have some

Can you expand on this? A providerID should only ever be set once GCP has accepted the instances create request, ie, we've told it to create a VM, it has started that process

pkg/cloud/gcp/actuators/machine/actuator.go
odvarkadaniel: That is the situation with the machine being OK and we change to zone to some nonsense. The machine still exists, but we only want to Fail it.

pkg/cloud/gcp/actuators/machine/actuator.go
odvarkadaniel: I didn't go deep into this, but I would expect this expression to be true `providerID == ""`, if the machine is not created properly, but that is not the case. So I just looked into differences in config of valid and invalid machines and came up with the node ref solution.

pkg/cloud/gcp/actuators/machine/actuator.go
JoelSpeed: Ok, so this is a fake exists, we haven't actually checked with the cloud provider, we just think it should exist?

Perhaps it's better not to deal with that and just handle all `invalidMachineConfiguration` scenarios here

pkg/cloud/gcp/actuators/machine/actuator.go
JoelSpeed: How did you simulate the machine not being created properly?

odvarkadaniel: 1. `oc get machineset <ms-name> -o yaml > invalid.yml`
2. Change zone, name and labels from `us-central-1a` to `us-central-1d`
3. `oc create -f invalid.yml`

pkg/cloud/gcp/actuators/machine/actuator.go
odvarkadaniel: I did it this way, because for the invalid machine, the phase is nil and we have to make sure it is failed. But if the machine was indeed valid before (exists) we want to fail it as well, but i don't want to assume what state is it in now.
IMO it is better to split the logic for used-to-be-valid/invalid machines.

pkg/cloud/gcp/actuators/machine/actuator.go
JoelSpeed: I just tried this with the current code from master, I get the following log messages
```
I1214 15:13:03.073223       1 controller.go:187] jspeed-test-kq9kx-worker-f-tcxst: reconciling Machine
I1214 15:13:03.073259       1 actuator.go:84] jspeed-test-kq9kx-worker-f-tcxst: Checking if machine exists
E1214 15:13:03.205859       1 controller.go:297] jspeed-test-kq9kx-worker-f-tcxst: failed to check if machine exists: jspeed-test-kq9kx-worker-f-tcxst: Machine does not exist
E1214 15:13:03.213819       1 controller.go:326]  "msg"="Reconciler error" "error"="jspeed-test-kq9kx-worker-f-tcxst: Machine does not exist" "controller"="machine-controller" "name"="jspeed-test-kq9kx-worker-f-tcxst" "namespace"="openshift-machine-api" "object"={"name":"jspeed-test-kq9kx-worker-f-tcxst","namespace":"openshift-machine-api"} "reconcileID"="5414d86b-1a74-4c76-9567-f06b1b1067b3"
```

The providerID is not set on the machine and the phase remains empty. So the providerID should be ok for invalid machines as far as I can tell

odvarkadaniel:
This new commit is mostly code structure changes. Logic stays the same as the last commit.
In this one, we made sure to check `providerID` instead of `NodeRef` of a machine. We unfortunately have to Fail the already existing machines ourselves, because worker machines don't Fail automatically as masters do.
Also had to move call for function `unregisterInstanceFromControlPlaneInstanceGroup` after we get to know if a machine is actually valid or not. If the machine is valid, we can continue as usual.

JoelSpeed:
> because worker machines don't Fail automatically as masters do.

The masters fail automatically but the workers don't? There should be no difference

odvarkadaniel:
> > because worker machines don't Fail automatically as masters do.
> 
> The masters fail automatically but the workers don't? There should be no difference

Yes. That's what I encountered multiple times.

JoelSpeed:
Any idea why? That's not behaviour I would expect to see

odvarkadaniel:
> Any idea why? That's not behaviour I would expect to see

I think I do. I worked it really badly last time.

If you have a running instance of master/worker machine and you edit the zone to some invalid zone. Both of the machines should be changed to "Failed". But if we create a new machine set with an invalid zone, a new worker machine, for example, gets created but is not failed by the controller.

Here we see that machine is Failed and won't get reconciled: https://github.com/openshift/machine-api-provider-gcp/blob/main/vendor/github.com/openshift/machine-api-operator/pkg/controller/machine/controller.go#L290-L293

Here we return an error, because the zone does not exist: https://github.com/openshift/machine-api-provider-gcp/blob/6148d6fb262205787f70cd2d2eea686c8288b3d2/pkg/cloud/gcp/actuators/machine/reconciler.go#L509-L511

`providerID` is nil and also `Phase` is nil, but since we return an error, we go to this block of code: https://github.com/openshift/machine-api-provider-gcp/blob/main/vendor/github.com/openshift/machine-api-operator/pkg/controller/machine/controller.go#L296-L310

But as we can see a little bit below, there is a piece of code that updates the `Phase` of the machine here: https://github.com/openshift/machine-api-provider-gcp/blob/main/vendor/github.com/openshift/machine-api-operator/pkg/controller/machine/controller.go#L350-L362
We never get to execute this, because of the err from function `r.actuator.Exists(ctx, m)`.

JoelSpeed:
I understand the flow you're talking through, but I still don't get why there's a difference between worker and master machines, they should be identical really as far as Machine API is concerned

odvarkadaniel:
> I understand the flow you're talking through, but I still don't get why there's a difference between worker and master machines, they should be identical really as far as Machine API is concerned

Yes, that is right. I thought I expressed that in my last comment.

JoelSpeed:
> Yes, that is right. I thought I expressed that in my last comment.

Ok I think I've got it, we used to have some logic in this PR that singled out control plane instances I think, we still have something there for instance groups, but that's existing code that's moved, got ya!

odvarkadaniel:
> > Yes, that is right. I thought I expressed that in my last comment.
> 
> Ok I think I've got it, we used to have some logic in this PR that singled out control plane instances I think, we still have something there for instance groups, but that's existing code that's moved, got ya!

Yes, exactly. I think with this version, every possible case of behavior is covered. I have also tested it on a GCP cluster and everything runs as expected.

pkg/cloud/gcp/actuators/machine/actuator.go
```diff
@@ -96,7 +97,31 @@ func (a *Actuator) Exists(ctx context.Context, machine *machinev1.Machine) (bool
...
+		// If the machine has, e.g. invalid zone, and we delete the invalid machinset,
+		// we want to set the machine to "Deleting" phase and return nil as error.
+		// We need the error to be nil so we can successfully delete the invalid machine.
+		if *machine.Status.Phase == "Deleting" {
+			return false, nil
+		}
+	}
+
+	return exists, err
```
JoelSpeed, line 124: So, I think we can refactor this a little to simplify it. One thing I'm uncomfortable is changing the meaning of the `exists` bool within `exists()`, so I think we should do the providerID check closer to where that information is being used, I think it makes it easier to follow in the long term
```suggestion
	exists, err := newReconciler(scope).exists()
	if !isInvalidMachineConfigurationError(err) {
		return exists, err
	}

	// If we got here, then the machine configuration is invalid.

	// If the machine has, e.g. invalid zone, and it has no phase set,
	// we want to set the machine to "Failed" phase so that we don't attempt
	// to create it again.
	if machine.Status.Phase == nil {
		machine.Status.Phase = pointer.String("Failed")
	}

	// If the machine has, e.g. invalid zone, and we delete the invalid machinset,
	// we want to set the machine to "Deleting" phase and return nil as error.
	// We need the error to be nil so we can successfully delete the invalid machine.
	// If the machine has a provider ID, then we believe something exists in the cloud.
	// So we must not all the deletion if the provider ID is set.
	if *machine.Status.Phase == "Deleting" && machine.Spec.ProviderID == nil {
		return false, nil
	}

	return false, err
```

Breaking this down, this first part retains the old behaviour and then explains that for the rest of this block the machine config is invalid
```
	exists, err := newReconciler(scope).exists()
	if !isInvalidMachineConfigurationError(err) {
		return exists, err
	}

	// If we got here, then the machine configuration is invalid.
```
Then, if the phase is nil, this means this is the first time we are seeing this Machine and it's been invalid from the start, so we can move it to failed. If the phase is nil, it definitely won't exist (unless something really bad has happened which is out of scope), so we don't need any other checks wrapping this.
```
	// If the machine has, e.g. invalid zone, and it has no phase set,
	// we want to set the machine to "Failed" phase so that we don't attempt
	// to create it again.
	if machine.Status.Phase == nil {
		machine.Status.Phase = pointer.String("Failed")
	}
```
Then, we have the caveat of is the Machine deleting. We override the error when the Machine is in the deleting phase and we don't see a providerID. I think it's important to do the providerID check here as it makes it very clear where and why we are doing this, with the additional comments as well.
```
	// If the machine has, e.g. invalid zone, and we delete the invalid machinset,
	// we want to set the machine to "Deleting" phase and return nil as error.
	// We need the error to be nil so we can successfully delete the invalid machine.
	// If the machine has a provider ID, then we believe something exists in the cloud.
	// So we must not all the deletion if the provider ID is set.
	if *machine.Status.Phase == "Deleting" && machine.Spec.ProviderID == nil {
		return false, nil
	}
```
Finally, return a false explicitly, we should always return false if we don't know, it's up to the caller to handle the error.
```
	return false, err
```

With this change, we then need a slight change to the other file too, I'll make suggestions for those as well

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -492,7 +503,13 @@ func (r *Reconciler) exists() (bool, error) {
 			// this error type bubbles back up to the machine-controller to allow
 			// us to delete machines that were never properly created due to
 			// invalid zone.
-			return false, machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("%s: Machine does not exist", r.machine.Name))
+
+			// If the machine has a node assigned, we know that it actually exists,
+			// but it also has an invalid configuration.
+			if r.machine.Spec.ProviderID != nil {
+				return true, machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("%s: Zone does not exist", r.providerSpec.Zone))
+			}
```
JoelSpeed, line 511: I don't think we want this, it creates a weird semantic around the `exists` bool returned here and I think that makes it more complicated at the function call sites
```suggestion
```

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -515,21 +532,27 @@ func (r *Reconciler) delete() error {
...
-	}
-
+	// Make sure that the machine exists.
+	// Also check that we have a machine with valid configuration.
 	exists, err := r.exists()
-	if err != nil {
+	if exists && isInvalidMachineConfigurationError(err) {
+		return fmt.Errorf("the machine %s has invalid configuration, but already exists, make the configuration of the machine valid for the deletion to be successful", r.machine.Name)
+	}
+	if err != nil && !isInvalidMachineConfigurationError(err) {
```
JoelSpeed, line 541: If the config is invalid and there's a providerID, we aren't going to proceed as we think there's already a machine present. We don't use `exists` because actually, we don't know if it does exist or not, I think this makes it more explicit.
```suggestion
	if isInvalidMachineConfigurationError(err) && machine.Spec.ProviderID != nil {
		return fmt.Errorf("the machine %s has invalid configuration, but already exists, make the configuration of the machine valid for the deletion to be successful", r.machine.Name)
	}
	if err != nil && !isInvalidMachineConfigurationError(err)  {
```

pkg/cloud/gcp/actuators/machine/actuator.go
odvarkadaniel: I think you are forgetting about machine that used to be valid and is now invalid. This case is not covered and such machine would be stuck in `running` even though it should be failed I think.

odvarkadaniel:
Every additional question resolved via Slack with @JoelSpeed.

odvarkadaniel:
Tested these changes on a GCP cluster, also checked the GCP dashboard that we don't leave any machines behind. Everything seems to be working just fine!

JoelSpeed:
Changes look good, are there any unit tests or anything we can look to extend to cover the changes in functionality?

odvarkadaniel:
> Changes look good, are there any unit tests or anything we can look to extend to cover the changes in functionality?

Added 2 test cases. I can't do any test cases with `actuator.Exists(ctx, machine)`, because there are no Events there.
So just these 2 are added. I think that it is enough for the basic functionality.

pkg/cloud/gcp/actuators/machine/actuator_test.go
```diff
@@ -210,6 +211,24 @@ func TestActuatorEvents(t *testing.T) {
...
 			event: "Deleted machine test",
 		},
+		{
+			name: "Delete machine with invalid machine config",
+			operation: func(actuator *Actuator, machine *machinev1.Machine) {
+				machine.Spec.ProviderID = nil
+				machine.Status.Phase = pointer.String("Deleting")
+				actuator.Delete(context.Background(), machine)
+			},
+			event: "test: reconciler failed to Delete machine: requeue in: 20s",
```
JoelSpeed, line 221: What makes this test invalid config?

odvarkadaniel: That it doesn't have a `providerID` and is in `Deleting` phase also. If the machine had a valid config, it would have a proper `providerID`.

PR #40, opened 2023-02-13
OCPCLOUD-1812: Update tooling & k8s 1.26

RadekManak:
Updates test tooling to be vendored instead of using shell script to download it. Also updates k8s to 1.26.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/openshift/release:golang-1.18 AS builder
+FROM registry.ci.openshift.org/openshift/release:golang-1.19 AS builder
```
lobziik, line 1: I wondering why the ART bot didn't update this?

lobziik (commented):
I'm a little bit curious about why go version was not updated by ART automation, but don't think that it's a reason for blocking this

/approve

Dockerfile
RadekManak: ART builds use Dockerfile.rhel. Not sure if this Dockerfile is used anywhere right now.

JoelSpeed:
/retest-required

Looks like the unit tests are suddenly taking longer to run and hitting a global timeout 🤔

PR #43, opened 2023-02-18
OCPBUGS-7366: New machine stuck in provisioning on VPC

odvarkadaniel:
New machine is stuck in `Provisioning` when we delete one of the zones from CPMS on GCP :  "The resource `<path-to-network-resource>` not found".

- [x] Propose a solution to the bug
- [x] Create unit test(s) covering this issue
- [x] Test this issue on running GCP cluster with custom VPC network

Edit: This got tested pre-merged by QE.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -663,15 +663,41 @@ func (r *Reconciler) ensureInstanceGroup(instanceGroupName string) error {
 	return nil
 }
 
+// ensureCorrectNetworkAndSubnetName makes sure that we are working with the correct network and subnet.
+// In a case of VPC, we have to look whether the expect name of the network and subnet resource
+// matches the one, that is actually up in the cluster.
+func (r *Reconciler) ensureCorrectNetworkAndSubnetName() (string, string, error) {
+	foundNetworkInterface := false
+	actualNetworkName := fmt.Sprintf("%s-network", r.machine.Labels[machinev1.MachineClusterIDLabel])
+	actualSubnetworkName := fmt.Sprintf("%s-%s-subnet", r.machine.Labels[machinev1.MachineClusterIDLabel], r.machineScope.machine.ObjectMeta.Labels[openshiftMachineRoleLabel])
+	for _, network := range r.providerSpec.NetworkInterfaces {
+		if network.Network == actualNetworkName && network.Subnetwork == actualSubnetworkName {
+			foundNetworkInterface = true
+			break
+		}
+	}
+	if !foundNetworkInterface {
+		actualNetworkName = r.providerSpec.NetworkInterfaces[0].Network
+		actualSubnetworkName = r.providerSpec.NetworkInterfaces[0].Subnetwork
+	}
+
+	return actualNetworkName, actualSubnetworkName, nil
+}
+
 // registerNewInstanceGroup registers an instance group when there is an instance
 // that is using that unkown instance group.
 func (r *Reconciler) registerNewInstanceGroup() error {
-	_, err := r.computeService.InstanceGroupInsert(r.projectID, r.providerSpec.Zone, &compute.InstanceGroup{
+	actualNetworkName, actualSubnetworkName, err := r.ensureCorrectNetworkAndSubnetName()
+	if err != nil {
+		return fmt.Errorf("failed to ensure the correct network name: %v", err)
+	}
+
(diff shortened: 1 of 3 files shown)


odvarkadaniel:
Hi @sunzhaohua2, could you please pre-merged test this? When I tried to reproduce this, I got totally different errors. Thank you!

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -663,15 +663,38 @@ func (r *Reconciler) ensureInstanceGroup(instanceGroupName string) error {
 	return nil
 }
 
+// ensureCorrectNetworkAndSubnetName makes sure that we are working with the correct network and subnet.
+// In a case of VPC, we have to look whether the expect name of the network and subnet resource
+// matches the one, that is actually up in the cluster.
+func (r *Reconciler) ensureCorrectNetworkAndSubnetName() (string, string, error) {
```
JoelSpeed, line 669: So this function, checks if any of the networks match the "expected" format based on our installer expectations, else we just return the network of the first interface? Seems reasonable, but why not always go for the first network interface? Any thoughts on when this might be different?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -663,15 +663,38 @@ func (r *Reconciler) ensureInstanceGroup(instanceGroupName string) error {
...
+	return actualNetworkName, actualSubnetworkName, nil
+}
+
 // registerNewInstanceGroup registers an instance group when there is an instance
 // that is using that unkown instance group.
 func (r *Reconciler) registerNewInstanceGroup() error {
-	_, err := r.computeService.InstanceGroupInsert(r.projectID, r.providerSpec.Zone, &compute.InstanceGroup{
+	actualNetworkName, actualSubnetworkName, err := r.ensureCorrectNetworkAndSubnetName()
+	if err != nil {
+		return fmt.Errorf("failed to ensure the correct network name: %v", err)
```
JoelSpeed, line 689: Nit, use %w for errors
```suggestion
		return fmt.Errorf("failed to ensure the correct network name: %w", err)
```

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -1164,3 +1164,83 @@ func TestRestartPolicyToBool(t *testing.T) {
...
+	}
+
+	for i, tc := range cases {
+		// Append the expected format of network and subnetwork according to GCP docs
+		if i == 1 {
+			r.providerSpec.NetworkInterfaces = append(r.providerSpec.NetworkInterfaces, &machinev1.GCPNetworkInterface{
+				Network:    "test-machine-1-network",
+				Subnetwork: "test-machine-1-test-machine-role-subnet",
+			})
+		}
```
JoelSpeed, line 1227: This is fragile and won't scale well, what happens if we want to include more cases later?

I would suggest instead to have a slice of network interfaces in your cases struct and set that within each test case, then the entire variation is included within the slice

I think we need several cases anyway:
* Single interface with unexpected values
* Single interface with expected values
* Two interfaces, expected in index 0
* Two interfaces, expected in index 1
Can you work on ensuring we have all of these cases?

pkg/cloud/gcp/actuators/machine/reconciler.go
odvarkadaniel: Depends what we would prefer. The thing you are suggesting makes sense as well.
I just wanted to split these 2 options: 1. VPC 2. Not VPC. I think the way it is now is also more explicit about what we expect the network and subnet to be on a regular cluster.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
odvarkadaniel: Good point. Will modify this.

odvarkadaniel: This won't be possible with the current implementation and I don't think that this is an issue.
This is also a good point for the current implementation IMO. If there is 20 networks and the one that we expect (the one that is always created by the installer) is on, let's say, 6th index, we don't want to just return 0th index of the `networkInterfaces`. I think that it would be better to just always put the machine in the basic network, unless the user decides otherwise by deleting this network resource - VPC use case.

odvarkadaniel:
/title OCPBUGS-7366: New machine stuck in provisioning on VPC

PR #54, opened 2023-06-27
CFE-684: Add user defined tags to the created gcp resource

bharath-b-rh:
PR has the changes for [enhancement](https://github.com/openshift/enhancements/pull/1217) proposed to support GCP tags in OCP, which requires machine-api-provider-gcp to add gcp userTags available in the status sub resource of infrastructure CR, to all the gcp resources created by the operator.


pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +162,118 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+	if len(userTags) == 0 && len(t.machineProviderSpecTags) == 0 {
+		klog.V(1).Infof("getResourceManagerTags: user-defined tags in Infrastructure and MachineProviderSpec is empty")
+		return nil, nil
+	}
+
+	mergeInfraProviderSpecTags(&userTags, t.machineProviderSpecTags)
+
+	if len(userTags) > 50 {
+		return nil, fmt.Errorf("maximum of 50 tags can be added to a compute instance, "+
+			"infrstructure.status.resourceTags and Machine.Spec.ProviderSpec.UserTags put "+
```
JoelSpeed, line 268: Please use the lower camelCase versions of the field names, they should not be capitalised

Also, Infrastructure is misspelt

go.mod
```diff
@@ -123,3 +131,5 @@ require (
 	sigs.k8s.io/kustomize/kyaml v0.14.3-0.20230601165947-6ce0bf390ce3 // indirect
 	sigs.k8s.io/structured-merge-diff/v4 v4.3.0 // indirect
 )
+
+replace github.com/openshift/api => ../../github.com/openshift/api
```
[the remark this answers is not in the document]
bharath-b-rh: Yes, it was for tests only and api package version will be updated once the API PR is merged.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels_test.go
```diff
@@ -252,3 +368,279 @@ func TestGetLabelsList(t *testing.T) {
...
+		{
+			name:               "providerSpec tags maxes allowed limit",
+			tagsFeatureEnabled: true,
+			getInfra: func() *configv1.Infrastructure {
+				infra := new(configv1.Infrastructure)
+				infraObj.DeepCopyInto(infra)
+				return infra
+			},
+			machineSpecTags: []machinev1.ResourceManagerTag{
+				{ParentID: "openshift", Key: "key1", Value: "value1"}, {ParentID: "openshift", Key: "key2", Value: "value2"},
```
theobarberbany, line 508: nit: storing these in a `const` or a lambda and referring to them may make readability better (and avoid some duplication)

pkg/cloud/gcp/actuators/util/gcp_tags_labels_test.go
```diff
@@ -252,3 +368,279 @@ func TestGetLabelsList(t *testing.T) {
...
+		},
+		{
+			name:               "user-defined tags maxes allowed limit",
+			tagsFeatureEnabled: true,
+			getInfra: func() *configv1.Infrastructure {
+				infra := new(configv1.Infrastructure)
+				infraObj.DeepCopyInto(infra)
+				infra.Status.PlatformStatus.GCP.ResourceTags = []configv1.GCPResourceTag{
+					{ParentID: "openshift", Key: "key1", Value: "value1"}, {ParentID: "openshift", Key: "key2", Value: "value2"},
+					{ParentID: "openshift", Key: "key3", Value: "value3"}, {ParentID: "openshift", Key: "key4", Value: "value4"},
```
theobarberbany, line 546: same as above! :)

theobarberbany (commented):
Looks good after a first pass! I have a few questions / nits :)

cmd/manager/main.go
```diff
@@ -143,12 +143,12 @@ func main() {
...
 	)
 	if err != nil {
-		klog.Fatalf("failed to create feature gate accessor: %w", err)
+		klog.Fatalf("failed to create feature gate accessor: %v", err)
 	}
 
 	featureGates, err := awaitEnabledFeatureGates(featureGateAccessor, 1*time.Minute)
 	if err != nil {
-		klog.Fatalf("failed to get feature gates: %w", err)
+		klog.Fatalf("failed to get feature gates: %v", err)
```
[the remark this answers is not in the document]
JoelSpeed: Errors should always be `%w` in format strings now

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -93,6 +96,12 @@ func newMachineScope(params machineScopeParams) (*machineScope, error) {
 		klog.V(1).Infof("status of %v feature is %t", configv1.FeatureGateGCPLabelsTags, gcpLabelsTagsFeature)
 	}
 
+	tagManager, err := util.NewTagManager(params.coreClient, gcpLabelsTagsFeature,
+		serviceAccountJSON, providerSpec.ResourceManagerTags)
+	if err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
JoelSpeed: InvalidMachineConfiguration only used for definite errors where we know the user input is wrong. What kind of errors are these? Could they be from user input or more likely to be transient?

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -442,13 +443,27 @@ func TestCreate(t *testing.T) {
 				Zone:                "test-zone",
 				MachineType:         "n2d-standard-4",
 				ConfidentialCompute: machinev1.ConfidentialComputePolicyEnabled,
+				ResourceManagerTags: []machinev1.ResourceManagerTag{
+					{ParentID: "openshift", Key: "key1", Value: "value1"},
```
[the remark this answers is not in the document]
JoelSpeed: +1 for readability

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -190,6 +190,14 @@ func (r *Reconciler) create() error {
 		},
 	}
 
+	userTags, err := r.resourceTagManager.GetResourceManagerTags(r.Context)
+	if err != nil {
+		return machinecontroller.CreateMachine("failed to fetch user-defined tags for %s: %v", r.machine.Name, err)
```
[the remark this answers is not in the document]
JoelSpeed: `%w` for errors please

cmd/manager/main.go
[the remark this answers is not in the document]
bharath-b-rh: Arguments to `klog.Fatalf ` are handled same as `fmt.Printf`, which doesn't support `%w` and `go vet` is reporting error for the same. 
```
cmd/manager/main.go:146:3: k8s.io/klog/v2.Fatalf does not support error-wrapping directive %w
cmd/manager/main.go:151:3: k8s.io/klog/v2.Fatalf does not support error-wrapping directive %w
```
Yeah, instead of `%v` we can use `%s`. I will make the change.

pkg/cloud/gcp/actuators/machine/machine_scope.go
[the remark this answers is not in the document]
bharath-b-rh: Currently, the method does not have an error returning scenario. I changed the code which was doing a bit more, than just mapping the args to internal model. I will update func definition to return just one value.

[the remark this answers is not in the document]
JoelSpeed: If it needs to return an error, it can return an error, we just need to decide how to handle that error

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -442,13 +443,27 @@ func TestCreate(t *testing.T) {
...
+		{
+			name: "failed to fetch resource manager tags",
+			providerSpec: &machinev1.GCPMachineProviderSpec{
+				Region: "test-region",
+				Zone:   "test-zone",
+				ResourceManagerTags: []machinev1.ResourceManagerTag{
+					{ParentID: "openshift", Key: "key2", Value: "value2"},
+				},
+			},
+			expectedError: errors.New("failed to fetch user-defined tags for : fail to fetch resource manager {openshift key2 value2} tag"),
```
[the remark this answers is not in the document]
bharath-b-rh: Error returned is from a fake method used for tests. Error returned by the actual code will contain the detailed error returned from google API and for which machine.
```
return machinecontroller.CreateMachine("failed to fetch user-defined tags for %s: %v", r.machine.Name, err)
```

pkg/cloud/gcp/actuators/machine/reconciler_test.go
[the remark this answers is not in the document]
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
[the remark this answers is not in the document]
bharath-b-rh: error message updated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels_test.go
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels_test.go
bharath-b-rh: suggestion incorporated.

cmd/manager/main.go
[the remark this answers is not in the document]
theobarberbany: Ah I did not realise this, my bad!

go.mod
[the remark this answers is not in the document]
bharath-b-rh: Updated `openshift/api` package version to latest.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +162,114 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+		tagValueList[value.Parent] = value.Name
+	}
+
+	return tagValueList, nil
+}
+
+// mergeInfraProviderSpecTags merges user-defined tags in Infrastructure.Status and
+// GCPMachineProviderSpec, with precedence given to those in GCPMachineProviderSpec
+// for new or updated tags. infraTags arg will contain the merged list.
+func mergeInfraProviderSpecTags(infraTags *[]machinev1.ResourceManagerTag, providerSpecTags []machinev1.ResourceManagerTag) {
```
JoelSpeed, line 221: I don't like that we are passing in and modifying tags here, would be better to return our own list that is fresh

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +162,114 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+	for _, pTag := range providerSpecTags {
+		appendTag := true
+		for i, iTag := range *infraTags {
+			if pTag.ParentID == iTag.ParentID && pTag.Key == iTag.Key && pTag.Value == iTag.Value {
+				appendTag = false
```
JoelSpeed, line 226: Just append directly?

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +162,114 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+				break
+			}
+			if pTag.ParentID == iTag.ParentID && pTag.Key == iTag.Key && pTag.Value != iTag.Value {
+				appendTag = false
+				(*infraTags)[i].Value = pTag.Value
```
JoelSpeed, line 231: I think there is a better way to do this. If you start with a copy of all of the providerSpecTags, then you'll be able to iterate over the infra tags and add any where the parentID and key are not found in the list, ignoring and not overwriting any values

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +162,114 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+			*infraTags = append(*infraTags, pTag)
+		}
+	}
+}
+
+// GetResourceManagerTags returns the merged list of user-defined tags in Infrastructure.Status
+// and GCPMachineProviderSpec to apply on the resources.
+func (t *tagManager) GetResourceManagerTags(ctx context.Context) (map[string]string, error) {
+	if !t.tagsFeatureEnabled {
+		klog.V(1).Infof("GCPLabelsTags featureGate is not enabled, skip adding tags")
```
JoelSpeed, line 245: I think the logging at the depth of code we are at should be higher than this, can you make this and the other logs a V3 please

pkg/cloud/gcp/actuators/util/gcp_tags_labels_test.go
```diff
@@ -14,20 +14,505 @@ limitations under the License.
...
+// tagSetIndexRange is for picking defined tags in the given index range
+// from testTagSets. low and high are both inclusive bounds.
+type tagSetIndexRange struct {
+	low  int
+	high int
+}
+
+// getTestTagSet returns the list of tags in the provided multiple ranges from the
+// testTagSets.
+func getTestTagSet[T configv1.GCPResourceTag | machinev1.ResourceManagerTag](tags *[]T, indexRange ...tagSetIndexRange) {
```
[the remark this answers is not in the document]
bharath-b-rh: Yeah, both are structs having the same exact member fields, and since both are populated from the common source, thought to use generics and cast it to the requested type.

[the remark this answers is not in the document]
bharath-b-rh: Yeah true, but I am using it as a way to find out the required type `configv1.GCPResourceTag | machinev1.ResourceManagerTag`.

pkg/cloud/gcp/actuators/util/gcp_tags_labels_test.go
```diff
@@ -252,3 +737,336 @@ func TestGetLabelsList(t *testing.T) {
...
+			}
+
+			clientBuilder := controllerfake.NewClientBuilder()
+			if infra != nil {
+				clientBuilder.WithObjects(infra)
+			}
+			fakeClient := clientBuilder.WithScheme(scheme).Build()
+
+			testTagManager := NewTagManager(fakeClient, tc.tagsFeatureEnabled, "", machineSpecTags)
+			testTagManager.(*tagManager).tagClientOptions = getTestTagClientOptions(server)
```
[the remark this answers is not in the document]
bharath-b-rh: I had tried mocking the interfaces, but came across some issues, sorry I am not able to recall. I will try again to find out the issues faced.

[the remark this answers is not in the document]
bharath-b-rh: Suggestion incorporated. Fake server dependency is removed and interfaces are mocked instead. Earlier I had issues with mocking the API defined error type, and recent version has provided an API for creating the required error.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +162,114 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+// status sub-resource of Infrastructure.
+func getInfraResourceTagsList(platformStatus *configv1.PlatformStatus) (tags []machinev1.ResourceManagerTag) {
+	if platformStatus != nil && platformStatus.GCP != nil && platformStatus.GCP.ResourceTags != nil {
+		tags = make([]machinev1.ResourceManagerTag, len(platformStatus.GCP.ResourceTags))
+		for i, tag := range platformStatus.GCP.ResourceTags {
+			tags[i] = machinev1.ResourceManagerTag(tag)
+		}
+	}
+	return
+}
```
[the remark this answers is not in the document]
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +162,114 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+	return []option.ClientOption{
+		option.WithCredentialsJSON([]byte(cred)),
+		option.WithEndpoint(fmt.Sprintf("https://%s", resourceManagerHostSubPath)),
+	}
+}
+
+// getTagValuesClient returns the client to be used for reading tag values to
+// the resources.
+func (t *tagManager) getTagValuesClient(ctx context.Context) (*tagValuesClient, error) {
+	client, err := rscmgr.NewTagValuesRESTClient(ctx, t.tagClientOptions...)
```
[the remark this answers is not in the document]
bharath-b-rh: TagManager interface was being used for mocking the high level API called during machine creation and a new interface for tag client is created and used for mocking in tests.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
bharath-b-rh: I have updated the logic, but `appendTag` flag is still used to skip the duplicate tags.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels_test.go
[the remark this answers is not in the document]
JoelSpeed: I wonder would it be simpler to have a conversion function and use an internal type, or use only one of these types in the code

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -90,9 +93,12 @@ func newMachineScope(params machineScopeParams) (*machineScope, error) {
 	gcpLabelsTagsFeature := false
 	if params.featureGates != nil {
 		gcpLabelsTagsFeature = params.featureGates.Enabled(configv1.FeatureGateGCPLabelsTags)
-		klog.V(1).Infof("status of %v feature is %t", configv1.FeatureGateGCPLabelsTags, gcpLabelsTagsFeature)
+		klog.V(3).Infof("status of %v feature is %t", configv1.FeatureGateGCPLabelsTags, gcpLabelsTagsFeature)
 	}
 
+	tagManager := util.NewTagManager(params.coreClient, util.NewTagValuesClient, serviceAccountJSON,
```
JoelSpeed, line 99: Does this function return nil if the feature is disabled? I would prefer if the switch logic (should we use this feature or not) were at the call site, rather than within the tags manager itself. It makes it hard to reason about when debugging.

The tag manager should be the production ready code, it shouldn't need to carry or know about the state of the feature.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +175,135 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+			}
+		}
+		return tags
+	}
+	return nil
+}
+
+// GetNamespacedTagValue is a wrapper for rscmgr.GetNamespacedTagValue() method.
+func (c *tagValuesClient) GetNamespacedTagValue(ctx context.Context, req *rscmgrpb.GetNamespacedTagValueRequest) (*rscmgrpb.TagValue, error) {
+	return c.TagValuesClient.GetNamespacedTagValue(ctx, req)
```
JoelSpeed, line 213: This is the interaction with GCP right? Is tag management part of their compute service or a separate service? We currently have all GCP API calls for compute service in https://github.com/openshift/machine-api-provider-gcp/blob/main/pkg/cloud/gcp/actuators/services/compute/computeservice.go, can we make an equivalent/follow the same pattern for the parts of this code that interact with google APIs, so that we have a clean separation of concern for what is logic to talk to google, vs our internal tag management code

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -190,6 +190,14 @@ func (r *Reconciler) create() error {
 		},
 	}
 
+	userTags, err := r.resourceTagManager.GetResourceManagerTags(r.Context)
```
JoelSpeed, line 193: This should be wrapped in:
```
if featureGateAccess.Enabled(configv1.GCPResourceTags) {
  // Do tag manager things
}
```
Not sure that's the correct feature gate name or not

pkg/cloud/gcp/actuators/util/gcp_tags_labels_test.go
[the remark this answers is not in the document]
bharath-b-rh: A common type is defined for storing the tags, will make use of it and convert to required type at caller. Tags are read from both `Infrastructure` and `GCPMachineProviderSpec` hence in tests both types are used for populating those object kinds, but in code it's converted to one type and same type is used for processing tags.

pkg/cloud/gcp/actuators/machine/machine_scope.go
bharath-b-rh: I will update to have the featuregate check at caller to invoke TagManager only when enabled.

bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
bharath-b-rh: suggestion incorporated.
`google.golang.org/api/cloudresourcemanager/v1` didn't have support for tag related APIs earlier, which has been updated in v3. Code updated to keep it inline with other implementation.

pkg/cloud/gcp/actuators/machine/reconciler.go
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/services/tags/tagservice.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// BuilderFuncType is function type for building gcp tag client.
+type BuilderFuncType func(ctx context.Context, serviceAccountJSON string) (TagService, error)
+
+// NewTagService return a new tagService.
+func NewTagService(ctx context.Context, serviceAccountJSON string) (TagService, error) {
+	service, err := tags.NewService(ctx, option.WithCredentialsJSON([]byte(serviceAccountJSON)))
+	if err != nil {
+		return nil, err
```
JoelSpeed, line 27: Would appreciate wrapping this error please
```suggestion
		return nil, fmt.Errorf("could not create new tag service: %w", err)
```

pkg/cloud/gcp/actuators/services/tags/tagservice_mock.go
```diff
@@ -0,0 +1,26 @@
+package tagservice
+
+import (
+	"context"
+
+	tags "google.golang.org/api/cloudresourcemanager/v3"
+)
+
+type MockTagService struct {
```
JoelSpeed, line 9: Exported types and functions should have godocs

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +130,100 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+			tags[i] = machinev1.ResourceManagerTag{
+				ParentID: tag.ParentID,
+				Key:      tag.Key,
+				Value:    tag.Value,
+			}
+		}
+		return tags
+	}
+	return nil
+}
```
JoelSpeed, line 149: I think this would be simpler to follow if we negated this and removed some of the indentation
```suggestion
	if platformStatus == nil || platformStatus.GCP == nil || platformStatus.GCP.ResourceTags == nil {
		return nil
	}
	
	tags := make([]machinev1.ResourceManagerTag, len(platformStatus.GCP.ResourceTags))
	
	for i, tag := range platformStatus.GCP.ResourceTags {
		tags[i] = machinev1.ResourceManagerTag{
			ParentID: tag.ParentID,
			Key:      tag.Key,
			Value:    tag.Value,
		}
	}
	
	return tags
	
}
```
Takes up more lines, but in general is a more accepted pattern

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +130,100 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+			var gErr *apierror.APIError
+			// google API returns StatusForbidden or StatusNotFound when the tag
+			// does not exist, since it could be because of permission issues
+			// or genuinely tag does not exist.
+			if errors.As(err, &gErr) && (gErr.HTTPCode() == http.StatusNotFound ||
+				gErr.HTTPCode() == http.StatusForbidden) {
+				klog.Errorf("does not have permission to access %s tag or tag does not exist", name)
+				continue
+			}
+			return nil, err
```
JoelSpeed, line 169: We intentionally fail if fetching a single tag fails? Is that a fatal error here? I guess this is only for temporary resource fetching errors right?

Can we wrap this with some context please?

pkg/cloud/gcp/actuators/services/tags/tagservice_mock.go
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/services/tags/tagservice.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+
+	return &tagService{
+		tagValuesService: tags.NewTagValuesService(service),
+	}, nil
+}
+
+func (t *tagService) GetNamespacedName(ctx context.Context, name string) (*tags.TagValue, error) {
```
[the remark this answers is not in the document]
bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
bharath-b-rh: We should be able to fetch details of every configured tag, since its required to be passed during Compute resource creation. Other option I think would be to maintain state in status.Condition for tag specific errors and retry adding tags which failed here after resource is created. 

I removed ignoring the permission denied or non-existent tag errors, I think even those should be left to user to rectify, correct?

User configured tags contain parentID, tagKey and tagValue short names, which will be used for deriving tag's namespaced name of the form `parentID/tagKey/tagValue` and is used for fetching tag's key and value names of the form `tagKeys/<UID>`, `tagValues/<UID>` as expected by the create compute API.

pkg/cloud/gcp/actuators/util/gcp_tags_labels.go
```diff
@@ -119,3 +127,97 @@ func GetLabelsList(userLabelsAllowed bool, client controllerclient.Client, machi
...
+	tagValueList := make(map[string]string, len(tagList))
+	for _, tag := range tagList {
+		name := fmt.Sprintf("%s/%s/%s", tag.ParentID, tag.Key, tag.Value)
+		value, err := tagService.GetNamespacedName(ctx, name)
+		if err != nil {
+			// fetching tag's metadata could fail due to errors like timeout, server
+			// internal errors, permission issues among others. Since tag's key and
+			// value names are required for binding tag to compute resource, will
+			// return error and retry during next reconciliation.
+			return nil, fmt.Errorf("failed to fetch %s tag details: %w", name, err)
```
JoelSpeed, line 163: I actually think we didn't want to drop the "this tag isn't found" section of this. If the error is transient, then retrying is good. If the error is, this tag isn't found, there's no point retrying. You can use an `InvalidMachineConfigurationError` to tell the machine controller to stop trying on permanent errors like that, I think that's what we want here

bharath-b-rh: suggestion incorporated.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -190,6 +191,21 @@ func (r *Reconciler) create() error {
...
 
+	var userTags map[string]string
+	if r.featureGates.Enabled(configv1.FeatureGateGCPLabelsTags) {
+		userTags, err = util.GetResourceManagerTags(r.Context, r.coreClient, r.tagService, r.providerSpec.ResourceManagerTags)
+		if err != nil {
+			switch err.(type) {
+			case *machinecontroller.MachineError:
+				return err
+			}
+			return machinecontroller.CreateMachine("failed to fetch user-defined tags for %s: %v", r.machine.Name, err)
```
JoelSpeed, line 202: This will be sufficient, if you using the wrapping semantic the core machine controller will pick up the invalid machine configuration, no need to use CreateMachine here
```suggestion
			return fmt.Errorf("failed to fetch user-defined tags for %s: %w", r.machine.Name, err)
```

bharath-b-rh: `MachineError` is used for just permission denied error case, wouldn't `CreateMachine` error be required for other errors where retry is required?

JoelSpeed: Retry happens no matter if there's an error or not. I checked the main code and it doesn't handle the `CreateMachine` specially in this case, it just unwraps any errors to see if somewhere in the stack it was an `InvalidMachineConfiguration`, so the suggestion I've made should have the intended effect

theobarberbany:
/lgtm 

Nice job!

elmiko (commented):
this mostly makes sense to me, just had a question about the unit tests

pkg/cloud/gcp/actuators/machine/machine_scope_test.go
```diff
@@ -215,6 +217,8 @@ func TestNewMachineScope(t *testing.T) {
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			gs := NewWithT(t)
+			tc.params.tagsClientBuilder = tagservice.NewMockTagServiceBuilder
+			tc.params.featureGates = featuregates.NewFeatureGate(nil, []configv1.FeatureGateName{configv1.FeatureGateGCPLabelsTags})
```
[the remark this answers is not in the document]
bharath-b-rh: Yeah, both scenarios are covered but not in same modules. Test cases added in actuator and machine_scope modules have the featuregate disabled and cases in reconciler have it enabled. And test cases added for gcp_tags modules is for enabled scenario.

PR #53, opened 2023-06-27
CFE-683 : Add user defined labels to the created gcp resource

bharath-b-rh:
PR has the changes for [enhancement](https://github.com/openshift/enhancements/pull/1217) proposed to support GCP labels in OCP, which requires machine-api-provider-gcp to add gcp userLabels available in the status sub resource of infrastructure CR, to all the gcp resources created by the operator.

PR #56, opened 2023-07-25
CFE-683 : Update openshift/api package to latest version

bharath-b-rh:
PR is for updating the openshift/api package to the latest version which contains the API changes for supporting GCP labels & tags made in https://github.com/openshift/api/pull/1445


JoelSpeed:
Updating the API is fine, but what about the actual implementation? We are waiting on support in cluster-api-provider-gcp as well right?

You'll need to update machine-api-operator for the webhooks as well btw

bharath-b-rh:
> Updating the API is fine, but what about the actual implementation? We are waiting on support in cluster-api-provider-gcp as well right?
> 
> You'll need to update machine-api-operator for the webhooks as well btw

cluster-api-provider-gcp already has support for adding labels. I have created an [issue](https://github.com/kubernetes-sigs/cluster-api-provider-gcp/issues/992) for supporting tags. But for TP tags for worker compute instances will be a limitation.

JoelSpeed:
What does this PR currently achieve? Does this controller already support labels as well? Does it need to be adding the labels from the infrastructure object now as well?

bharath-b-rh:
With this PR we will update the version of openshift/api which support resourceLabels configuration in Infrastructure API. Controller already has the functionality for adding those labels in Machine and MachineSet to Compute Instance, https://github.com/openshift/machine-api-provider-gcp/pull/53 requires the pkg update for merging user-defined labels in Infrastructure with the ProviderSpec labels.

PR #60, opened 2023-10-04
MIXEDARCH-258: Set the capacity annotation for arch-aware autoscale from zero

aleskandro:
This PR adds a facility to support arch-aware autoscale from zero in multiarch compute clusters.
Since the gcloud API does not provide the architecture information from a give machine type, we define a map of pairs `(prefix, architecture)`. If a known pair is available, the labels' capacity annotation is set to the appropriate architecture. Otherwise, the capacity labels annotation will set to `kubernetes.io/arch=amd64`

Today, GCP provides `arm64` machine types with the `t2a-` prefix. The drawback of this implementation is the need to maintain the map in future when new machine types could be added.

cc @Prashanth684

I'm not sure this path can be viable. I found other code in this actuator that leverages checks on the machine type prefixes to perform some actions (for example, https://github.com/aleskandro/machine-api-provider-gcp/blob/41599ce77bfb7689dab017af80e7b8afd9c0a890/pkg/cloud/gcp/actuators/machine/reconciler.go#L133-L149). Opening the PR to discuss together wheter possible or to look at other solutions, if any.

Could you have a look at [this diff](https://github.com/openshift/machine-api-provider-gcp/pull/60/commits/96bfd90c0eae727fb85ba47cfb7ef6ba0b34c770)?

### DIFF
--- a/pkg/cloud/gcp/actuators/machineset/controller.go
+++ b/pkg/cloud/gcp/actuators/machineset/controller.go
@@ -23,6 +23,7 @@ import (
 	"github.com/go-logr/logr"
 	machinev1 "github.com/openshift/api/machine/v1beta1"
 	mapierrors "github.com/openshift/machine-api-operator/pkg/controller/machine"
+	mapiutil "github.com/openshift/machine-api-operator/pkg/util"
 	computeservice "github.com/openshift/machine-api-provider-gcp/pkg/cloud/gcp/actuators/services/compute"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -40,6 +41,7 @@ const (
 	cpuKey    = "machine.openshift.io/vCPU"
 	memoryKey = "machine.openshift.io/memoryMb"
 	gpuKey    = "machine.openshift.io/GPU"
+	labelsKey = "capacity.cluster-autoscaler.kubernetes.io/labels"
 )
 
 // Reconciler reconciles machineSets.
@@ -168,6 +170,11 @@ func (r *Reconciler) reconcile(machineSet *machinev1.MachineSet) (ctrl.Result, e
 		machineSet.Annotations[gpuKey] = strconv.FormatInt(0, 10)
 	}
 
+	// We guarantee that any existing labels provided via the capacity annotations are preserved.
+	// See https://github.com/kubernetes/autoscaler/pull/5382 and https://github.com/kubernetes/autoscaler/pull/5697
+	machineSet.Annotations[labelsKey] = mapiutil.MergeCommaSeparatedKeyValuePairs(
+		fmt.Sprintf("kubernetes.io/arch=%s", util.CPUArchitecture(providerConfig.MachineType)),
+		machineSet.Annotations[labelsKey])
 	return ctrl.Result{}, nil
 }
 
--- a/pkg/cloud/gcp/actuators/util/gcp_machine_architecture.go
+++ b/pkg/cloud/gcp/actuators/util/gcp_machine_architecture.go
@@ -1,0 +1,41 @@
+/*
+Copyright The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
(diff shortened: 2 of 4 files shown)


aleskandro:
/cc @elmiko

as we were talking of this some time ago

Prashanth684:
A while ago azure used to have instance type list which was copied from the [autoscaler](https://github.com/openshift/kubernetes-autoscaler/blob/master/cluster-autoscaler/cloudprovider/azure/azure_instance_types.go):
https://github.com/openshift/machine-api-provider-azure/pull/17
These were mostly generated by the az cli commands. But since then we don't use this. 

Could we take a similar approach or should we think about changing the API to include the architecture information which might be bigger in scope and possibly involve upstream discussion.

pkg/cloud/gcp/actuators/util/gcp_machine_architecture.go
```diff
@@ -0,0 +1,41 @@
...
+const (
+	// ArchitectureAmd64 is the normalized architecture name for amd64.
+	ArchitectureAmd64 NormalizedArch = "amd64"
+	// ArchitectureArm64 is the normalized architecture name for arm64.
+	ArchitectureArm64 NormalizedArch = "arm64"
+)
+
+// machineTypePrefixArchitectureMap contains a map of (machineTypePrefix, architecture) tuples
+var machineTypePrefixArchitectureMap = map[string]NormalizedArch{
+	"t2a": ArchitectureArm64,
```
Prashanth684, line 29: better to maintain a curated file with a list of all instance types and fetch the info from there rather than do this?

aleskandro: Do you mean a file to unmarshal at need or to add more families to this map?

aleskandro:
> A while ago azure used to have instance type list which was copied from the [autoscaler](https://github.com/openshift/kubernetes-autoscaler/blob/master/cluster-autoscaler/cloudprovider/azure/azure_instance_types.go): [openshift/machine-api-provider-azure#17](https://github.com/openshift/machine-api-provider-azure/pull/17) These were mostly generated by the az cli commands. But since then we don't use this.
> 
> Could we take a similar approach or should we think about changing the API to include the architecture information which might be bigger in scope and possibly involve upstream discussion.


Compared to the Azure case, we can use just the prefixes of the machine types for GCP, which are not a lot.

However, there is no command to my knowledge that allow us to automatically build a "family -> architecture" map.


> Could we take a similar approach or should we think about changing the API to include the architecture information which might be bigger in scope and possibly involve upstream discussion.

Do you mean in the gcloud SDK?

pkg/cloud/gcp/actuators/util/gcp_machine_architecture.go
Prashanth684: i thought a file with all the families possible from the cmdline. But if that is not possible, we might have to live with this.

go.mod
```diff
@@ -5,27 +5,27 @@ go 1.18
...
-	k8s.io/client-go v0.27.2
+	github.com/onsi/ginkgo/v2 v2.12.1
+	github.com/onsi/gomega v1.28.0
+	github.com/openshift/api v0.0.0-20230928105710-23b54c280f99
+	github.com/openshift/client-go v0.0.0-20230926161409-848405da69e1
+	github.com/openshift/library-go v0.0.0-20230927113136-405c34317fa4
+	github.com/openshift/machine-api-operator v0.2.1-0.20230929171041-2cc7fcf262f3
+	golang.org/x/oauth2 v0.8.0
+	google.golang.org/api v0.114.0
+	k8s.io/api v0.28.2
```
JoelSpeed, line 16: You probably need to rebase as we updated all the deps to 1.28 in another PR recently

pkg/cloud/gcp/actuators/machineset/controller.go
```diff
@@ -40,6 +41,7 @@ const (
 	cpuKey    = "machine.openshift.io/vCPU"
 	memoryKey = "machine.openshift.io/memoryMb"
 	gpuKey    = "machine.openshift.io/GPU"
+	labelsKey = "capacity.cluster-autoscaler.kubernetes.io/labels"
```
JoelSpeed, line 44: @elmiko I think you should probably review this

pkg/cloud/gcp/actuators/util/gcp_machine_architecture.go
JoelSpeed: How frequently are new instance families likely to be added such that we are going to need to update this? I guess there's no API that GCP exposes currently that gives you a sense of the architecture of an instance type or family?

aleskandro: No, no exposed API to gather the architecture of a given machine type unfortunately.

As for the frequency, I'm not sure, the first arm64 instances have been released this year.

elmiko (commented):
thanks for your patience @aleskandro 

/approve

PR #71, opened 2023-11-21
OCPBUGS-2117: Add extra check to termination handler

elmiko:
This change adds an extra check the termination handler so that it can try one more time after having the context cancelled.  It is needed to fix a race condition where the termination signal can be missed and the machine is not properly marked for removal.

This change also fixes a few deprecated calls in the termination handler.

### DIFF
--- a/pkg/termination/termination.go
+++ b/pkg/termination/termination.go
@@ -3,7 +3,7 @@ package termination
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"net/url"
 	"sync"
@@ -95,17 +95,17 @@ func (h *handler) run(ctx context.Context) error {
 	logger := h.log.WithValues("node", h.nodeName)
 	logger.V(1).Info("Monitoring node termination")
 
-	if err := wait.PollImmediateUntil(h.pollInterval, func() (bool, error) {
+	if err := wait.PollUntilContextCancel(ctx, h.pollInterval, true, func(_ context.Context) (bool, error) {
 		terminated, err := h.checkTerminationEndpoint()
 		if !terminated {
 			logger.V(2).Info("Instance not marked for termination")
 		}
 		return terminated, err
-	}, ctx.Done()); err != nil {
+	}); err != nil {
 		return fmt.Errorf("error polling termination endpoint: %w", err)
 	}
 
-	// We might arrive here due to the context being canceled before we have gotten
+	// We might arrive here due to the context being cancelled before we have gotten
 	// a clean signal from the termination endpoint, we check once more.
 	if terminated, err := h.checkTerminationEndpoint(); err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


pkg/termination/termination.go
```diff
@@ -95,61 +95,76 @@ func (h *handler) run(ctx context.Context) error {
 	logger := h.log.WithValues("node", h.nodeName)
 	logger.V(1).Info("Monitoring node termination")
 
-	if err := wait.PollImmediateUntil(h.pollInterval, func() (bool, error) {
-		req, err := http.NewRequest("GET", h.pollURL.String(), nil)
-		if err != nil {
-			return false, fmt.Errorf("could not create request %q: %w", h.pollURL.String(), err)
+	if err := wait.PollUntilContextCancel(ctx, h.pollInterval, true, func(_ context.Context) (bool, error) {
```
JoelSpeed, line 98: What happens when the context cancels on this function?

pkg/termination/termination.go
```diff
@@ -95,61 +95,76 @@ func (h *handler) run(ctx context.Context) error {
...
 	// Will only get here if the termination endpoint returned FALSE
 	logger.V(1).Info("Instance marked for termination, marking Node for deletion")
 
 	// Try every second to mark the node for termination up to a 30 second timeout.
 	// This should help to prevent intermittent errors and ensure we don't end up in crash loop backoff.
 	markCtx, cancel := context.WithTimeout(ctx, 30*time.Second)
 	defer cancel()
-	if err := wait.PollImmediateUntil(time.Second, func() (bool, error) {
+	if err := wait.PollUntilContextCancel(markCtx, time.Second, true, func(_ context.Context) (bool, error) {
 		if err := h.markNodeForDeletion(ctx); err != nil {
```
JoelSpeed, line 124: This needs to use a different context I think, I expect the parent `ctx` may already be cancelled at this point

JoelSpeed (commented):
/hold

I think the action of marking might fail here as the original context is likely to be cancelled already before we use it on l124

pkg/termination/termination.go
elmiko: i misread the comments for this function, i had thought it did not return the error on cancel but looking at the code it does. we need an extra `else if` to catch the cancel i think.

pkg/termination/termination.go
elmiko: agreed, i'll have to think this one through a little more. i suppose we could create a new background context when we go into the code path where the context is already cancelled.

JoelSpeed: `markCtx` will also be dead by this point so that one on l121 will need some attention too

pkg/termination/termination.go
elmiko: i've updated this to check for a canceled context in the else so that it will continue to the rest of the function if the context was canceled.

elmiko:
not sure why the unit failed on that one run, logs look weird

elmiko:
@JoelSpeed thanks, i'm good to let this go for qe testing. i would love to confirm this in a way that made me feel happy but i think i'm wrestling with racing the cloud to catch the termination.

PR #89, opened 2024-08-20
OCPCLOUD-2565: Featuregate implementation

theobarberbany:
This change: 

- Updates the MAO, to bring in new changes to the `Machine` Reconciler.
- Reads in feature gates from flags, passing them to the reconciler. 

Running the binary without passing the --feature-gates flag will result in the feature gates initialising as follows: 

```
I0812 17:36:32.555187    1120 main.go:154] Initializing feature gates: AllAlpha=true|false (ALPHA - default=false), AllBeta=true|false (BETA - default=false), MachineAPIMigration=true|false (ALPHA - default=false)
I0812 17:36:32.555275    1120 main.go:159] FeatureGateMachineAPIMigration initialised: false
```

Note that any feature gates that are GA will not be printed in the help text, but can still be set by flags.


JoelSpeed (commented):
Looks good
/hold until library-go dependent PR is merged though

PR #94, opened 2024-10-21
CORS-3739: Add c4a instance as arm type

barbacbd:
** The c4a instance type needs to be added for other projects that import this package and its validation. Added the c4a instance type as an ARM type to the dictionary for lookup.


makentenza:
@barbacbd @JoelSpeed, as a reminder, we'll need to backport support for this to OCP 4.17.z

PR #98, opened 2024-12-11
OCPBUGS-45923: update a2 gpu detection logic to be dynamic

elmiko:
this change updates the way that a2 instance types are validated for quota. gcp has added a new type of a2 instance category named "ultragpu". the ultragpu series has a different type of gpu, "nvidia-a100-80gb", which requires a new constant be applied to the quota.

to make this workflow more modular in the future, it is also being adjusted to dynamically discover the gpu type during the instance quota validation.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -49,6 +49,8 @@ func newReconciler(scope *machineScope) *Reconciler {
 }
 
 var (
+	// the keys have been sourced from https://cloud.google.com/compute/docs/gpus/
+	// the values have been sourced from https://github.com/googleapis/google-api-go-client/blob/main/compute/v1/compute-gen.go
 	supportedGpuTypes = map[string]string{
 		"nvidia-tesla-k80":  "NVIDIA_K80_GPUS",
 		"nvidia-tesla-p100": "NVIDIA_P100_GPUS",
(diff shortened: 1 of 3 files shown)


elmiko:
@JoelSpeed we talked about doing a fix that was holistic and added the a3 instance types to the quota checking, but after a discussion in the afternoon my time we were talking about the timing around getting the full change in. i proposed the idea of fixing this for a2 instances since a3's were already working and in the worst case scenario the errors are no worse than before.

this patch is a stepping stone to adding full quota detection for a3 as well. but it allows us to get a more refined patch out that doesn't necessarily add new conditional logic for a3 instances.

elmiko:
i'm just doing some cleanups on the code

elmiko:
also, i did manual cherry picks to get the 4.18 and 4.17 branches queued up, the bugs are linked

JoelSpeed:
This gets us something working while we work on a better long term fix

/lgtm
/approve

@elmiko Do you have a separate tracker for the longer term fix?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -144,17 +140,21 @@ func (r *Reconciler) validateGuestAccelerators() error {
 	}
 	a2MachineFamily, n1MachineFamily := r.computeService.GPUCompatibleMachineTypesList(r.providerSpec.ProjectID, r.providerSpec.Zone, r.Context)
 	machineType := r.providerSpec.MachineType
-	switch {
-	case a2MachineFamily[machineType] != 0:
+	if gpuInfo, ok := a2MachineFamily[machineType]; ok {
 		// a2 family machine - has fixed type and count of GPUs
```
nrb, line 144: Is this comment out of date now? Realize this is merged, but maybe we update it if/when we touch the code again for a3 instances.

PR #101, opened 2024-12-12
OCPBUGS-43531: Refactor exists() to handle gcp API change

RadekManak:
GCP started responding with 400 for invalid zone instead of 404. This PR updates the exits() to handle the change.

It also removes the zone get API call and instead determines invalid zone or projectID from the error message of the instance get call.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -530,29 +530,28 @@ func (r *Reconciler) exists() (bool, error) {
 	if err := validateMachine(*r.machine, *r.providerSpec); err != nil {
 		return false, fmt.Errorf("failed validating machine provider spec: %v", err)
 	}
-	zone := r.providerSpec.Zone
-	// Need to verify that our project/zone exists before checking machine, as
-	// invalid project/zone produces same 404 error as no machine.
-	if err := r.validateZone(); err != nil {
-		if isNotFoundError(err) {
-			// this error type bubbles back up to the machine-controller to allow
-			// us to delete machines that were never properly created due to
-			// invalid zone.
 
+	_, err := r.computeService.InstancesGet(r.projectID, r.providerSpec.Zone, r.machine.Name)
+	if err != nil {
+		// InvalidMachineConfiguration error type bubbles back up to the machine-controller to allow
+		// us to delete machines that were never properly created due to
+		// invalid zone or project.
+		if isProjectNotFoundError(err, r.projectID) {
+			return false, machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("%s: Project does not exist", r.projectID))
+		}
+
+		if isInvalidZone(err) {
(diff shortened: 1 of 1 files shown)


RadekManak:
/retest-required
I am seeing some failures from GCP that eventually recovered. I don't think this PR is causing it. Let's retest.

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -608,6 +607,22 @@ func isNotFoundError(err error) bool {
 	return false
 }
 
+func isProjectNotFoundError(err error, projectID string) bool {
+	switch t := err.(type) {
+	case *googleapi.Error:
+		return t.Code == 404 && strings.Contains(t.Message, fmt.Sprintf("The resource 'projects/%s' was not found", projectID))
```
elmiko, line 613: is there any documentation around this error message, it makes me a little nervous because what if google changes the error body?

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -608,6 +607,22 @@ func isNotFoundError(err error) bool {
...
+	}
+	return false
+}
+
+func isInvalidZone(err error) bool {
+	switch t := err.(type) {
+	case *googleapi.Error:
+		return t.Code == 400 && strings.Contains(t.Message, "Invalid value for field 'zone':")
```
elmiko, line 621: similar question here

elmiko (commented):
this makes sense to me, but i have couple questions.

also, can we remove `validateZone` ?

pkg/cloud/gcp/actuators/machine/reconciler.go
RadekManak: I think https://google.aip.dev/193#:~:text=Otherwise%2C%20the%20content%20of%20Status.message%20must%20be%20stable%2C%20providing%20the%20same%20text%20with%20the%20same%20request%2Dspecific%20information. applies here as there is no ErrorInfo in the response.
```
{
  "error": {
    "code": 404,
    "message": "The resource 'projects/openshift-gce-develd' was not found",
    "errors": [
      {
        "message": "The resource 'projects/openshift-gce-develd' was not found",
        "domain": "global",
        "reason": "notFound"
      }
    ],
    "status": "NOT_FOUND"
  }
}
```

I have updated the function to look for 'projects/%s' and notFound reason. If google changes this I am eating my Fedora.

pkg/cloud/gcp/actuators/machine/reconciler.go
RadekManak: see above

RadekManak:
```
{
  "error": {
    "code": 400,
    "message": "Invalid value for field 'zone': 'us-east4-ax'. Unknown zone.",
    "errors": [
      {
        "message": "Invalid value for field 'zone': 'us-east4-ax'. Unknown zone.",
        "domain": "global",
        "reason": "invalid"
      }
    ]
  }
}
```

JoelSpeed:
If google changes the format, this will regress, but as Radek says, we can eat fedoras when they do

/lgtm
/approve

PR #102, opened 2024-12-13
OCPBUGS-46387: add a3 instance types to gpu quota validation logic

elmiko:
this change expands the quota detection logic to include a3 instance types in addition to a2.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler.go
@@ -52,13 +52,15 @@ var (
 	// the keys have been sourced from https://cloud.google.com/compute/docs/gpus/
 	// the values have been sourced from https://github.com/googleapis/google-api-go-client/blob/main/compute/v1/compute-gen.go
 	supportedGpuTypes = map[string]string{
-		"nvidia-tesla-k80":  "NVIDIA_K80_GPUS",
-		"nvidia-tesla-p100": "NVIDIA_P100_GPUS",
-		"nvidia-tesla-v100": "NVIDIA_V100_GPUS",
-		"nvidia-tesla-a100": "NVIDIA_A100_GPUS",
-		"nvidia-tesla-p4":   "NVIDIA_P4_GPUS",
-		"nvidia-tesla-t4":   "NVIDIA_T4_GPUS",
-		"nvidia-a100-80gb":  "NVIDIA_A100_80GB_GPUS",
+		"nvidia-tesla-k80":      "NVIDIA_K80_GPUS",
+		"nvidia-tesla-p100":     "NVIDIA_P100_GPUS",
+		"nvidia-tesla-v100":     "NVIDIA_V100_GPUS",
+		"nvidia-tesla-a100":     "NVIDIA_A100_GPUS",
+		"nvidia-tesla-p4":       "NVIDIA_P4_GPUS",
+		"nvidia-tesla-t4":       "NVIDIA_T4_GPUS",
+		"nvidia-a100-80gb":      "NVIDIA_A100_80GB_GPUS",
+		"nvidia-h100-mega-80gb": "NVIDIA_H100_GPUS",
+		"nvidia-h100-80gb":      "NVIDIA_H100_GPUS",
 	}
 )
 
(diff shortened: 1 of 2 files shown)


pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -127,20 +127,23 @@ func (r *Reconciler) checkQuota(guestAccelerators []machinev1.GCPGPUConfig) erro
...
+	if len(r.providerSpec.GPUs) > 0 && (strings.HasPrefix(r.providerSpec.MachineType, "a2-") || strings.HasPrefix(r.providerSpec.MachineType, "a3-")) {
+		return machinecontroller.InvalidMachineConfiguration("A2 and A3 Machine types have pre-attached guest accelerators. Adding additional guest accelerators is not supported")
 	}
-	if !strings.HasPrefix(r.providerSpec.MachineType, "n1-") && !strings.HasPrefix(r.providerSpec.MachineType, "a2-") {
-		return machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("MachineType %s does not support accelerators. Only A2 and N1 machine type families support guest acceleartors.", r.providerSpec.MachineType))
+	if !strings.HasPrefix(r.providerSpec.MachineType, "n1-") && !strings.HasPrefix(r.providerSpec.MachineType, "a2-") && !strings.HasPrefix(r.providerSpec.MachineType, "a3-") {
+		return machinecontroller.InvalidMachineConfiguration(fmt.Sprintf("MachineType %s does not support accelerators. Only A2, A3 and N1 machine type families support guest acceleartors.", r.providerSpec.MachineType))
```
michaelryanmcneill, line 142: Typo: acceleartors should be accelerators

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -127,20 +127,23 @@ func (r *Reconciler) checkQuota(guestAccelerators []machinev1.GCPGPUConfig) erro
 }
 
 func (r *Reconciler) validateGuestAccelerators() error {
-	// Note(elmiko) this is known to have an error in that non a2 instances with GPUs (eg a3 types) will bypass this check, which is fine for now.
-	if len(r.providerSpec.GPUs) == 0 && !strings.HasPrefix(r.providerSpec.MachineType, "a2-") {
+	// Note, this logic will reject machines that might are not a2 or a3, or have GPUs specified. This means that future
```
nrb, line 130: > Note, this logic will reject machines that might are not a2 or a3

Not exactly sure what this is meant to say. Maybe omitting `might`?

elmiko: oh yeah, good catch

JoelSpeed:
/approve
/lgtm
/label acknowledge-critical-fixes-only

/hold 

Want to check with QE first before merging

pkg/cloud/gcp/actuators/machine/reconciler.go
```diff
@@ -52,13 +52,15 @@ var (
...
-		"nvidia-tesla-t4":   "NVIDIA_T4_GPUS",
-		"nvidia-a100-80gb":  "NVIDIA_A100_80GB_GPUS",
+		"nvidia-tesla-k80":      "NVIDIA_K80_GPUS",
+		"nvidia-tesla-p100":     "NVIDIA_P100_GPUS",
+		"nvidia-tesla-v100":     "NVIDIA_V100_GPUS",
+		"nvidia-tesla-a100":     "NVIDIA_A100_GPUS",
+		"nvidia-tesla-p4":       "NVIDIA_P4_GPUS",
+		"nvidia-tesla-t4":       "NVIDIA_T4_GPUS",
+		"nvidia-a100-80gb":      "NVIDIA_A100_80GB_GPUS",
+		"nvidia-h100-mega-80gb": "NVIDIA_H100_GPUS",
```
jianli-wei, line 62: How about "NVIDIA_H100_MEGA_GPUS"?

JoelSpeed: I don't think we have a choice here, if you read the comment above, this is mapping the keys from the documentation to the values in the SDK

elmiko: we could put `NVIDIA_H100_MEGA_GPUS` here, but since it isn't used in the AWS SDK, it would never match. unfortunately we are limited by what this version of the SDK will allow us to detect from the calls to AWS.

JoelSpeed:
/hold cancel

I think this is good to go

PR #108, opened 2025-01-28
OCPBUGS-17079, OCPBUGS-44671: Disable shielded VMs for non-UEFI disks

theobarberbany:
This PR builds upon https://github.com/openshift/machine-api-provider-gcp/pull/80

- Abstracts the check to a helper function/utility
- Adds proper parsing for all supported input formats of disk images.

It also adds an additional check in the machine reconciler's `create()` method, to ensure creation of new machinesets non-uefi-compatible won't fail (the higher level machineset check may have not run or updated the spec when creation happens)


pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// checkImageUEFICompatible retrieves the image and checks its GuestOSFeatures for UEFI support.
+func checkImageUEFICompatible(gceService computeservice.GCPComputeService, project, image string) (bool, error) {
+	img, err := gceService.ImageGet(project, image)
+	if err != nil {
+		return false, fmt.Errorf("unable to retrieve image %q in project %q: %w", image, project, err)
+	}
+	for _, feat := range img.GuestOsFeatures {
+		if strings.Contains(feat.Type, UEFICompatible) {
```
JoelSpeed, line 110: Why contains and not equal?

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check.go
```diff
@@ -0,0 +1,131 @@
...
+			klog.Infof("Image %q is UEFI compatible: %v", image, img.GuestOsFeatures)
```
JoelSpeed, line 111: At how many layers do we log this? I thought I saw a similar log at the call site 🤔

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check.go
```diff
@@ -0,0 +1,131 @@
...
+	if !strings.Contains(imageStr, "/") {
+		return &imageReference{
+			Project:  providerProject,
+			Image:    imageStr,
+			IsFamily: false,
+		}, nil
+	}
+
+	// Check if the image string contains "projects/".
+	if !strings.Contains(imageStr, "projects/") {
```
nrb, line 65: Are you doing `Contains` here so that it supports full URLs? You [mentioned on Slack](https://redhat-internal.slack.com/archives/GE2HQ9QP4/p1737654701440449?thread_ts=1737479970.161129&cid=GE2HQ9QP4) that full URLs are valid. If that's still correct, maybe add that to the list of supported formats and a test case for it.

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check.go
nrb: I thought because it's passed in as a comma-separated string (https://cloud.google.com/compute/docs/images/create-custom#guest-os-features), but the [http API docs](https://cloud.google.com/compute/docs/reference/rest/v1/images/get) for getting an image show that it is indeed an array of strings containing values from an enum.

nrb (commented):
Thanks for this, it's much more thorough than my old fix. Added a couple comments as there may be a corner case not covered.

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check.go
theobarberbany: This is just late night refactor brain 🤦 Should be equal!

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check.go
theobarberbany: This was more to verify the format of the inputted disk image string. I think the split should strip everything before `/projects`, keeping full URLs supported, but if we don't have any `/projects`, but do have a `/` we have an unexpected input which may break the parsing below

theobarberbany: I also want to add tests for this function so it's really clear how it's expected to function, so will put a full URL in the cases for that.

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check.go
theobarberbany: Yep, makes sense - there were debuggy I left in. 

Have moved up, and changed the log that the check is beginning to be only visible at `v=3` as I think it'll be noise otherwise.

pkg/cloud/gcp/actuators/machineset/controller.go
```diff
@@ -175,6 +175,31 @@ func (r *Reconciler) reconcile(machineSet *machinev1.MachineSet) (ctrl.Result, e
 	machineSet.Annotations[labelsKey] = mapiutil.MergeCommaSeparatedKeyValuePairs(
 		fmt.Sprintf("kubernetes.io/arch=%s", util.CPUArchitecture(providerConfig.MachineType)),
 		machineSet.Annotations[labelsKey])
+
+	// From OCP 4.13, we enable Shielded Instance Config by default. However, in
+	// some cases the boot disk is not UEFI compatible and instance creation will
```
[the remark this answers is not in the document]
theobarberbany: I'm not 100% that I'm following here, could be me!: you suggest: Mention that OCP images don't support UEFI pre 4.13?

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check_test.go
```diff
@@ -0,0 +1,339 @@
...
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	machinev1 "github.com/openshift/api/machine/v1beta1"
+	computeservice "github.com/openshift/machine-api-provider-gcp/pkg/cloud/gcp/actuators/services/compute"
+	"github.com/openshift/machine-api-provider-gcp/pkg/cloud/gcp/actuators/util"
+)
+
+var _ = Describe("IsUEFICompatible", func() {
```
JoelSpeed, line 12: This suite would be better written as a table case rather than using `JustBeforeEach` I think, wanna give that a go?

pkg/cloud/gcp/actuators/machineset/controller.go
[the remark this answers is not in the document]
nrb: Yeah.

It's not inaccurate to say that `ShieldedInstanceConfig` being true was the default in 4.13, but it's also not complete; 4.12 and under did not have the field at all. IMO it would have been better to introduce the field for a version or two, then make the enabled state default, but that didn't happen.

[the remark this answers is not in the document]
theobarberbany: Ahh I see, I follow now. I was getting confused as some <=4.12 disk images are UEFI compatible (from my own digging).

Will update.

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check_test.go
```diff
@@ -0,0 +1,212 @@
...
+	var (
+		computeService computeservice.GCPComputeService
+		providerSpec   *machinev1.GCPMachineProviderSpec
+
+		compatible bool
+		err        error
+	)
```
JoelSpeed, line 19: Are these needed still? Are they being shared with the other set of tests at the end?

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check_test.go
```diff
@@ -0,0 +1,212 @@
...
+	})
+
+	type standardImageInput struct {
+		boot                 bool
+		image                string
+		projectID            string
+		zone                 string
+		expectedErrSubstring string
+		compatible           bool
+	}
```
JoelSpeed, line 33: If you added a `modifyProviderSpec func(*machinev1.GCPProviderSpec)`, option here, you could get all the tests in the one pattern, or, you could even just provider the disks directly `disks []*machinev1.GCPDisk` and construct that in each entry?

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check_test.go
```diff
@@ -0,0 +1,212 @@
...
+		if in.expectedErrSubstring != "" {
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring(in.expectedErrSubstring))
+		} else {
+			Expect(err).ToNot(HaveOccurred())
+		}
+		Expect(compatible).To(Equal(in.compatible))
+	}
+
+	DescribeTable("Standard image references in format projects/{project}/global/images/{image}",
```
JoelSpeed, line 61: Oh nice breaking this up into different groupings, makes sense

miyadav:
tested this functionality looks good.

pkg/cloud/gcp/actuators/util/gcp_uefi_disk_check_test.go
```diff
@@ -0,0 +1,206 @@
...
+		providerSpecBuilder machinev1builder.GCPProviderSpecBuilder
+		providerSpec        *machinev1.GCPMachineProviderSpec
+
+		compatible bool
+		err        error
+	)
+
+	BeforeEach(func() {
+		_, computeService = computeservice.NewComputeServiceMock()
+		providerSpecBuilder = machinev1builder.GCPProviderSpec()
```
JoelSpeed, line 25: You may as well move this into the table func right? And then you can get rid of all of the vars above?

PR #111, opened 2025-03-06
CORS-3906: Update MAPI GCP Provider to use custom GCP endpoints

barbacbd:
** Vendor update. Update openshift/api to use a version that contains the custom endpoint information.

** When creating the compute service, determine if there is a custom endpoint set in the infrastructure platform status.
** When creating the tagging service, determine fi there is a custom endpoint set in the infrastructure platform status.
** Add a function template to pass to the actuator and machine scope parameters for looking up the custom endpoints. During tests a Mock function is passed that will do nothing and return no error. During the actual execution the real lookup function is passed. When no function is set the default will be to use the function that will look through the infrastructure to find the custom endpoints.
** Update machine scope tests to set a mock function.
** Update actuator tests to set a mock function

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/actuator_test.go
+++ b/pkg/cloud/gcp/actuators/machine/actuator_test.go
@@ -320,6 +320,7 @@ func TestActuatorEvents(t *testing.T) {
 				ComputeClientBuilder: computeservice.MockBuilderFuncType,
 				TagsClientBuilder:    tagservice.NewMockTagServiceBuilder,
 				FeatureGates:         gate,
+				EndpointLookup:       util.MockGCPEndpointLookup,
 			}
 
 			actuator := NewActuator(params)
@@ -424,6 +425,7 @@ func TestActuatorExists(t *testing.T) {
 				ComputeClientBuilder: computeservice.MockBuilderFuncType,
 				TagsClientBuilder:    tagservice.NewMockTagServiceBuilder,
 				FeatureGates:         gate,
+				EndpointLookup:       util.MockGCPEndpointLookup,
 			}
 
 			actuator := NewActuator(params)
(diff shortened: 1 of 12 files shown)


barbacbd:
It appears that the failing unit test is also failing on my main branch too. It shouldn't be an architecture specific issue, but let's see how it does here.

JoelSpeed:
How does this work with CAPI? I assume we won't be patching CAPI to lookup endpoints, so is it through some configuration in the CAPI APIs?

barbacbd:
> How does this work with CAPI? I assume we won't be patching CAPI to lookup endpoints, so is it through some configuration in the CAPI APIs?

CAPI in this case the GCP provider was already updated. We use their Cluster Spec to send this information over to CAPG.

JoelSpeed:
Ok that makes sense, and for the CAPI tech preview right now, I think we have a controller that generates the GCP Cluster Spec, but we don't have anything (AFAIK CC @damdo) that would maintain certain config within it, like these endpoints.

To avoid a feature gap in the future, we either need to build that logic now, or create a card to track it.

Dam, do you see that as extending the existing InfraCluster controllers, or something separate that enforces certain fields based on the Infrastructure status as source?

pkg/cloud/gcp/actuators/machine/machine_scope.go
```diff
@@ -85,12 +89,33 @@ func newMachineScope(params machineScopeParams) (*machineScope, error) {
 		}
 	}
 
-	computeService, err := params.computeClientBuilder(serviceAccountJSON)
+	var computeEndpoint *configv1.GCPServiceEndpoint = nil
+	var tagsEndpoint *configv1.GCPServiceEndpoint = nil
+	if sets.New(params.featureGates.KnownFeatures()...).Has(string(apifeatures.FeatureGateGCPCustomAPIEndpoints)) {
```
JoelSpeed, line 94: You don't need to check if the feature is known, if the feature is not known, then something has gone very wrong

pkg/cloud/gcp/actuators/machineset/controller.go
```diff
@@ -215,7 +221,17 @@ func (r *Reconciler) getRealGCPService(namespace string, providerConfig machinev
 		return nil, err
 	}
 
-	computeService, err := computeservice.NewComputeService(serviceAccountJSON)
+	var endpoint *configv1.GCPServiceEndpoint = nil
+	if sets.New(r.FeatureGates.KnownFeatures()...).Has(string(apifeatures.FeatureGateGCPCustomAPIEndpoints)) {
```
JoelSpeed, line 225: Don't check if it's known

pkg/cloud/gcp/actuators/machine/actuator_test.go
```diff
@@ -311,7 +311,7 @@ func TestActuatorEvents(t *testing.T) {
 				return k8sClient.Get(context.Background(), machineKey, machine)
 			}
 			gs.Eventually(getMachine, timeout).Should(Succeed())
-
+
```
JoelSpeed, line 314: Hmm?

PR #113, opened 2025-04-11
CORS-3906: Update custom endpoint configuration

barbacbd:
If a custom service endpoint has been sent in for a service, use it during service creation. When this was originally completed, the option.WithEndpoint Google cloud option was used. This is the offically accepted way that is recommended by Google. However, it was discovered that when an endpoint formatted as https://someendpoint... is used, the client was producing 404 errors. If this same endpoint was used but set with BasePath = endpoint the correct endpoint was used and traffic was observed going to the endpoint.

This feature is behind the "GCPCustomAPIEndpoints" featuregate.


barbacbd:
the installer team will correctly format the data so that this change is not necessary.

PR #114, opened 2025-04-16
OCPBUGS-46387: allow instances with unknown quotas to be created

elmiko:
this change relaxes some of the quota checking logic so that if a
specific metric is not detecting in the list of quotas for a region, the
controller will emit a log warning but allow the creation of the
instance.

this is being introduced to help solve issues where the google apis do
not return regional information for some attached gpus, such as
NVIDIA_H100_GPU.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/reconciler_test.go
+++ b/pkg/cloud/gcp/actuators/machine/reconciler_test.go
@@ -27,14 +27,16 @@ import (
 
 func TestCreate(t *testing.T) {
 	cases := []struct {
-		name                string
-		labels              map[string]string
-		providerSpec        *machinev1.GCPMachineProviderSpec
-		expectedCondition   *metav1.Condition
-		secret              *corev1.Secret
-		mockInstancesInsert func(project string, zone string, instance *compute.Instance) (*compute.Operation, error)
-		validateInstance    func(t *testing.T, instance *compute.Instance)
-		expectedError       error
+		name                              string
+		labels                            map[string]string
+		providerSpec                      *machinev1.GCPMachineProviderSpec
+		expectedCondition                 *metav1.Condition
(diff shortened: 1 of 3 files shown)


elmiko:
i've updated the logic here in line with conversations we've had internally. we will allow the creation of machine types which have quota metrics which are not returned from the gcp endpoints.

a warning log message has been added detailing the instance type, gpu type, region, and zone of the unknown quota, and an advisory that instance creation may fail.

i've also added more mocking to the unit tests so that we can exercise the various logic branches.

nrb:
/override ci/prow/okd-scos-e2e-aws-ovn

This change is for GCP, not AWS.

pkg/cloud/gcp/actuators/machine/reconciler_test.go
```diff
@@ -175,6 +177,107 @@ func TestCreate(t *testing.T) {
...
+					"a2-highgpu-4g": computeservice.GpuInfo{
+						Type:  "nvidia-a100-80gb",
+						Count: 1,
+					},
+				}
+				return gpuInfo, compatibleMachineType
+			},
+			mockRegionGet: func(project string, region string) (*compute.Region, error) {
+				var computeQuota = &compute.Quota{
+					Limit:  1.0,
```
nrb, line 205: It looks like the limit is the only that's changing between the implementations of `mockRegionGet`; could we simplify the test struct to only include the limit and then have a shared method?

elmiko: sure, i can change it. i was following the previous pattern for injecting mock functions into the test structure.

nrb: Discussed at standup - the code as-is works and covers more cases.

PR #138, opened 2026-01-09
OCPBUGS-105414: Add the N4A machine type

barbacbd:
The N4A machine type is a new arm based machine type that GCP has released. The machine type should be valid when creating machines with the machine-api.


damdo (commented):
/lgtm

Thanks!

PR #160, opened 2026-04-21
OCPBUGS-82191: Wait for RUNNING before providerID; retry stockout

RadekManak:
GCP can accept InstancesInsert, show a VM in PROVISIONING or STAGING, then fail the insert and delete the instance (`ZONE_RESOURCE_POOL_EXHAUSTED*`).

The controller treated that as created: it wrote providerID, addresses, and MachineCreated=True. When the VM vanished, MAO set Failed with `can't find created instance`. The GCP error never appeared on the Machine. Create never ran again.

## What this PR does

**Report the insert-op error.** List the latest matching zone insert operation (ignore ops older than Machine.CreationTimestamp minus 2 minutes). If it is DONE with a provider error, put that on MachineCreated=False. Mixed error lists are terminal unless every non-empty code is pool-exhausted.

**Do not mark the Machine created until the VM is RUNNING.** Delay providerID, addresses, and MachineCreated=True. If we write them early, MAO Faileds and Create never runs, so the error we just learned to report is unused.

**On-demand pool-exhausted: retry this Machine for 30 minutes.** The clock starts at the first `MachineCreated=False` with Reason `MachineCreationCapacityFailed`, not at an earlier 5xx False. Flat 20s requeue. After 30 minutes on Create, InvalidMachineConfiguration. A later 500 does not overwrite that Reason or reset the clock.

**Spot/preemptible stockout: do not re-insert on the same object.** InvalidMachineConfiguration immediately. MachineSet replacement is the retry.

**Quota and other non-capacity insert failures** fail immediately on Create (`QUOTA_EXCEEDED` / HTTP 403 `quotaExceeded`, bad image, mixed lists).

**TERMINATED or STOPPED before providerID:** provision never reached RUNNING. Set providerID (and addresses if NICs exist), MachineCreated=True, delete the leftover VM, requeue. MAO Faileds once Exists is false.

**TERMINATED or STOPPED after providerID:** the VM ran (or was marked created). Do not delete. Requeue and wait for RUNNING so stop/start does not Failed the Machine. STOPPING / STAGING / PROVISIONING / REPAIRING still wait. Update never returns InvalidMachineConfiguration.

```mermaid
flowchart TD
  start[Reconcile] --> exists{Instance Exists?}

  exists -->|no Create| listOp[List insert ops after Machine.CreationTimestamp minus 2m]
  listOp --> cls{Op class}
  cls -->|in flight| rq1[Requeue]
  cls -->|pool-exhausted DONE, on-demand| recFalse[MachineCreated=False Reason CapacityFailed]
  recFalse --> dead{30m since that Reason?}
  dead -->|no| insert[InstancesInsert]
  dead -->|yes| termC[InvalidMachineConfiguration]
  cls -->|pool-exhausted DONE, Spot or preemptible| termC
  cls -->|other DONE error including mixed or quota| termC
  cls -->|success DONE, Get 404| insert
  cls -->|absent / stale| insert
  insert --> afterCreate[Get instance]
  afterCreate --> waitOrSet

  exists -->|yes Update| waitOrSet[Get instance]
  waitOrSet --> st{Status}
  st -->|RUNNING| markers[Set providerID, addresses, MachineCreated=True]
  markers --> maoP[MAO: phase Provisioned]
  maoP --> node{nodeRef set?}
  node -->|yes| maoR[MAO: phase Running]
  node -->|no| rq2[Requeue until Node joins]
  st -->|TERMINATED or STOPPED, no providerID| termDel[Set markers, delete leftover VM, requeue]
  termDel --> maoF[Next pass Exists=false: MAO Failed]
  st -->|TERMINATED or STOPPED, providerID set| rqStop[Requeue, do not delete]
  st -->|STOPPING / STAGING / PROVISIONING / REPAIRING| waitOp[Classify insert op, requeue, no markers]
  waitOp -->|pool-exhausted| recU[MachineCreated=False, requeue]
  recU --> noteU[Do not insert. Do not return InvalidMachineConfiguration.]
```

## Testing

Unit tests at InstancesGet / InstancesInsert / InstancesDelete / ZoneOperationsList: no providerID before RUNNING, pool-exhausted requeue and 30m Create deadline from the capacity Reason, 5xx False does not start that clock, 500 does not reset it, quota and mixed errors terminal on Create, Spot/preemptible no re-insert, stale insert ops ignored, TERMINATED/STOPPED before providerID markers plus delete, stop/start after providerID does not delete.

Live, we verified the Failed trap on the RUNNING gate. Stockout insert ops were proven earlier (`ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS` on vanished VMs). Quota text on MachineCreated=False was seen when we still retried quota; this PR fails quota immediately on Create.

## Context & Patterns

**MAO only Faileds InvalidMachineConfiguration on Create.** Update always requeues. A GCP TERMINATED instance is powered off and can be started again. We only delete that leftover when providerID was never set. After providerID, stop/start must not Failed the Machine.

**Quota is not stockout.** Pool-exhausted is HTTP 503 / try again later. Quota is 403 / raise the project limit. Mixed lists follow the stricter class.

**The 30m epoch is a Reason, not a new condition type.** Kubernetes does not move LastTransitionTime when Status stays False. We set Reason `MachineCreationCapacityFailed` and LastTransitionTime on the first stockout so an older 5xx False cannot expire the window on the first pool-exhausted insert.

**Spot retry is a new Machine, not a second insert.** Capacity is not guaranteed. MachineSet replacement is the retry.

### DIFF
--- a/pkg/cloud/gcp/actuators/machine/capacity.go
+++ b/pkg/cloud/gcp/actuators/machine/capacity.go
@@ -1,0 +1,99 @@
+package machine
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	machinev1 "github.com/openshift/api/machine/v1beta1"
+	machinecontroller "github.com/openshift/machine-api-operator/pkg/controller/machine"
+	"google.golang.org/api/compute/v1"
+	"google.golang.org/api/googleapi"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	capacityRetryDeadline = 30 * time.Minute
+)
+
+// Pool exhausted. Retry these; they are not a bad MachineSpec.
+// https://cloud.google.com/compute/docs/reference/rest/v1/errors
+var capacityErrorCodes = sets.NewString(
+	"ZONE_RESOURCE_POOL_EXHAUSTED",
+	"ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS",
+	"REGION_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS",
+	"RESOURCE_POOL_EXHAUSTED",
+)
+
+func isCapacityOperationError(op *compute.Operation) bool {
+	if op == nil || op.Error == nil {
+		return false
+	}
+	for _, e := range op.Error.Errors {
+		if e != nil && capacityErrorCodes.Has(e.Code) {
+			return true
+		}
+	}
+	return false
+}
+
+func isCapacityAPIError(err error) bool {
+	var googleError *googleapi.Error
+	if !errors.As(err, &googleError) {
+		return false
+	}
+	for _, item := range googleError.Errors {
+		if capacityErrorCodes.Has(item.Reason) {
+			return true
+		}
+	}
+	return false
+}
+
+func messageHasCapacityErrorCode(msg string) bool {
+	for _, code := range capacityErrorCodes.List() {
+		if strings.Contains(msg, code) {
+			return true
+		}
+	}
+	return false
+}
+
+// Keep the 30m create deadline from MachineCreated=False when list missed the insert op.
+func (r *Reconciler) capacityErrorFromCondition() (error, bool) {
+	cond := findCondition(r.providerStatus.Conditions, string(machinev1.MachineCreated))
+	if cond == nil || cond.Status != metav1.ConditionFalse {
+		return nil, false
+	}
+	if !messageHasCapacityErrorCode(cond.Message) {
+		return nil, false
+	}
+	return fmt.Errorf("%s", cond.Message), true
+}
+
+func (r *Reconciler) createCapacityMayInsert(err error) error {
+	r.recordFailedInstanceCreate(err)
+	cond := findCondition(r.providerStatus.Conditions, string(machinev1.MachineCreated))
+	if cond == nil || cond.Status != metav1.ConditionFalse || cond.LastTransitionTime.IsZero() {
+		return nil
+	}
+	if time.Since(cond.LastTransitionTime.Time) >= capacityRetryDeadline {
+		return machinecontroller.InvalidMachineConfiguration("capacity exhausted after %s: %v", capacityRetryDeadline, err)
+	}
+	return nil
+}
+
+func (r *Reconciler) requeueCapacityOnCreate(err error) error {
+	if deadlineErr := r.createCapacityMayInsert(err); deadlineErr != nil {
+		return deadlineErr
+	}
+	return &machinecontroller.RequeueAfterError{RequeueAfter: requeueAfterSeconds * time.Second}
+}
+
+// MAO ignores InvalidMachineConfiguration on Update, so this only requeues.
+func (r *Reconciler) requeueCapacityOnUpdate(err error) error {
+	r.recordFailedInstanceCreate(err)
+	return &machinecontroller.RequeueAfterError{RequeueAfter: requeueAfterSeconds * time.Second}
+}
--- a/pkg/cloud/gcp/actuators/machine/capacity_test.go
+++ b/pkg/cloud/gcp/actuators/machine/capacity_test.go
@@ -1,0 +1,481 @@
+package machine
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
+	configv1 "github.com/openshift/api/config/v1"
+	machinev1 "github.com/openshift/api/machine/v1beta1"
+	machinecontroller "github.com/openshift/machine-api-operator/pkg/controller/machine"
+	computeservice "github.com/openshift/machine-api-provider-gcp/pkg/cloud/gcp/actuators/services/compute"
+	tagservice "github.com/openshift/machine-api-provider-gcp/pkg/cloud/gcp/actuators/services/tags"
+	compute "google.golang.org/api/compute/v1"
+	googleapi "google.golang.org/api/googleapi"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/utils/ptr"
+	controllerfake "sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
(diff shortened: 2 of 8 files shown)


newtonheath:
@RadekManak can we get help getting this reviewed please - its a release blocker

stbenjam:
## Disposition

**REQUEST_CHANGES** — the core direction is sound, but four runtime-confirmed edge cases can still leave Machines failed or permanently stuck.

## Specialist Findings

<details>
<summary>Bugs — 2 BLOCKING</summary>

- Historical insert operations can fail a newly recreated Machine.
- The capacity deadline can inherit an unrelated failure's timestamp.

</details>

<details>
<summary>Adversarial — 4 BLOCKING</summary>

- Confirmed the two functional findings above.
- A Spot VM preempted before the first observed `RUNNING` state can remain unprovisioned forever.
- Mixed terminal and capacity errors are classified as retryable capacity failures.

</details>

<details>
<summary>Architecture — 1 BLOCKING, 2 SUGGESTIONS</summary>

- Operation lookup is not scoped to the current Machine lifetime.
- Capacity retry state depends on parsing human-readable condition text.
- `ZoneOperationsList` becomes a 20-second per-Machine polling primitive for visible, non-`RUNNING` instances.

</details>

<details>
<summary>QA — 4 SUGGESTIONS</summary>

- Add two-reconcile coverage for asynchronous capacity failure followed by an eventually consistent empty operation list.
- Cover an in-flight operation list miss followed by an insert conflict.
- Lock down the capacity deadline's transition-time invariant.
- Cover Update-side list errors and bound operation-list call counts.

</details>

<details>
<summary>Security, consistency, and documentation — no findings</summary>

Provider errors are reduced to code/message, no dependency or credential handling changes were introduced, repository conventions are followed, and existing documentation does not contradict the new behavior.

</details>

## Panel Synthesis

The deadline remains anchored across successive capacity messages, but it can inherit the timestamp of an earlier unrelated failure. Operation-list rediscovery is defensible without an API change, but results must be scoped to the current Machine generation. Delaying Provisioned is generally sound; visible terminal instances, particularly default-`STOP` Spot VMs, require explicit recovery.

The two-reconcile asynchronous-capacity-failure → empty-operation-list regression test is still missing. Current tests pre-seed the condition instead of exercising that sequence.

## Required Actions

1. **Scope operations to the current Machine lifetime** — `pkg/cloud/gcp/actuators/machine/insertops.go:11` matches only project, zone, operation type, and instance target link. A terminal operation from an older same-name Machine is selected by `reconcilePriorInsertAttempt`, which returns `InvalidMachineConfiguration` without attempting `InstancesInsert`. GCP retains completed operations for [at least 1 hour and up to 14 days](https://docs.cloud.google.com/compute/docs/instances/viewing-compute-operations#operation-retention-period). At minimum, reject operations predating the current Machine's `CreationTimestamp`, with an allowance for clock skew, and add a same-name recreation test.

2. **Start the 30-minute deadline at the first capacity failure** — `pkg/cloud/gcp/actuators/machine/capacity.go:76` uses the generic `MachineCreated=False` transition time. If unrelated retryable failures established that condition 31 minutes earlier, the first capacity error becomes terminal immediately. Track a capacity-specific retry epoch or explicitly reset it when entering the capacity-failure class, while preserving it across successive capacity failures.

3. **Handle terminal pre-provisioning instance states** — `pkg/cloud/gcp/actuators/machine/reconciler.go:551` requeues every visible non-`RUNNING` instance. Spot inserts leave `InstanceTerminationAction` unset, and GCP's default is [`STOP`](https://docs.cloud.google.com/compute/docs/instances/spot), so early preemption leaves a visible `TERMINATED` instance. `Exists` remains true, Update requeues indefinitely, and providerID/addresses/`MachineCreated=True` are never published. Add a bounded delete/recreate or terminal-failure path, with regression coverage.

4. **Make terminal errors dominate mixed error sets** — `pkg/cloud/gcp/actuators/machine/insertops.go:122` treats an operation as capacity-retryable if any error entry has a capacity code. Both `QUOTA_EXCEEDED + ZONE_RESOURCE_POOL_EXHAUSTED` orderings classify as capacity; the synchronous API classifier behaves the same way. Retry only when all actionable entries are recognized capacity errors, and surface all relevant codes.

<details>
<summary>Confirmed reproducer results</summary>

- **Stale operation:** an `INVALID_IMAGE` operation from 24 hours before the current Machine was selected; `InstancesInsert` was called zero times and the new Machine received `InvalidMachineConfiguration`.
- **Capacity clock:** a 31-minute-old non-capacity `MachineCreated=False` condition caused the first capacity error to return terminal immediately; the old timestamp remained unchanged.
- **Spot termination:** create plus four Update reconciles all returned a 20-second requeue; instance state remained `TERMINATED`, with providerID and addresses unset.
- **Mixed errors:** both error orderings returned `insertOperationCapacityFailed`; both synchronous orderings were also classified as capacity errors.

</details>

## Optional Follow-ups

- Add the asynchronous capacity-failure → empty-list two-reconcile regression test.
- Add the in-flight list miss → repeated insert → HTTP 409 reconciliation test.
- Document and bound `ZoneOperationsList` QPS for large MachineSets. GCP recommends polling the returned operation with `get` or `wait`; `wait` can [reduce QPS](https://docs.cloud.google.com/compute/docs/api/how-tos/api-requests-responses#polling-operations).
- Replace capacity-code parsing from condition messages with structured state.

## Stats

- 7 specialists; 13 raw findings; 4 deduplicated BLOCKING findings.
- 4/4 BLOCKING runtime reproducers confirmed.
- Changed-path tests and `git diff --check` passed; upstream unit and verification checks passed at the reviewed head.

<sub>Generated by the deep-review skill</sub>

RadekManak:
/hold
I am sceptical about the latest fixes around terminated. This got way to complex because of AI review.

RadekManak:
/hold cancel
Fixed the issues from review. Machine in TERMITATED state can still be restarted and finish boot. It should not reach this state unless something else stops it outside of controller logic. Leaving them requing will allow them to rejoin if they are also powered on outside of OpenShift. 

For SPOT I don't think the stockout retry is correct. The current behavior is for spot to go Failed and be replaced by another Machine and potentialy another MachineSet with instance type that has more capacity. I disabled the stockout retry for SPOT type Machines.

RadekManak:
/override ci/prow/e2e-gcp-ovn
Unrelated flaky test failure.

